PR #18, opened 2026-06-06
docs: conflict-driven learning design + go/no-go measurement

micheldumontier:
## What

Project-kickoff design for CDCL-style **conflict-driven nogood learning** on the hypertableau wedge — the only remaining lever for wine's disjunction-branching interdependence (per `docs/tableau-perf-scoping-2026-06-06.md`). **Design + measurement only; no engine code.**

## Go/no-go measurement (temporarily instrumented `solve` on wine, now reverted)

| Finding | Value | Implication |
|---|---|---|
| backjumping fires | **0 of 1.07M clashes** | every clash depends on the current decision → learning is genuinely the **only** lever (measured, not inferred — confirms the scoping verdict) |
| disjunction decisions on successors | **96%** (476k succ vs 17.9k root) | root-only learning is theater |
| decisions on deterministic nodes (`birth_deps=EMPTY`) | **~17.5%** | the other 82.5% are branch-created successors → need **canonical-by-creation-provenance** node identity to learn from them |

**Consequence:** there is **no de-risked on-ramp**. The hard, soundness-critical core (provenance-based canonical node identity) is unavoidable from the first useful increment.

## Soundness argument (rewritten per review)

Nogood soundness rests on (1) **monotonicity** of the learnable clashes (`A⊓¬A`, `⊥`, disjoint-present are only added) and (2) **overflow-exclusion** — the non-monotone rules (`≤n` merge, `≠`, NN) all set `clash_deps = DepSet::ALL`, so restricting learning to non-overflow clash-deps excludes exactly the unsound cases. Over-counting deps is safe (weaker nogood); never tighten.

## Honest verdict

Viable-in-principle but a **dedicated multi-week, soundness-critical effort with uncertain payoff** (nogood transfer on a branch-created-successor-dominated tree is unmeasured) targeting an **outlier** (the corpus already converges). The plan (Inc 0: provenance identity, gated byte-identical → Inc 1: record+match nogoods, gated FP=0 + differential fuzz) is in the doc. **Recommend starting fresh from Inc 0 in a devoted session; stop if Inc 1 shows weak wine `restores/branches` reduction.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #19, opened 2026-06-06
feat(tableau): conflict-driven nogood learning (Inc 1) — sound, weak; foundation for 1-UIP [DO NOT MERGE]

micheldumontier:
**Draft / do-not-merge.** Sound, default-OFF conflict-driven nogood learning + the measured finding that the *simple* form is too weak. Kept as the **foundation for a future 1-UIP CDCL effort**, not for merge to `main` (a default-OFF 13% feature is maintenance cost without 1-UIP). Companion to the design in #18.

## Built (behind `RUSTDL_HYPER_LEARNING`, default OFF)
- Canonical node identity (interned by `(parent_canon, role, qualifier, decision-path, disamb)`).
- Decision stack + interned decision ids; nogood store.
- Record non-overflow clash dep-sets as nogoods; prune a disjunct when active decisions ⊇ a nogood. Overflow clashes (`≤n`/`≠`/NN — non-monotone) excluded for soundness.

## Soundness — validated empirically (the key positive)
Corpus closure-diff with learning ON is **byte-identical to OFF**: FP=0, every MISSED unchanged across alehif/galen/notgalen/shoiq-knowledge/ore-10908/ore-15672/sio/wine. Confirms the monotonicity + overflow-exclusion soundness argument on real SROIQ/nominal/cardinality inputs.

## Payoff — weak (legitimate stop, not a bug)
wine: `recorded=1015` nogoods, `pruned=157733` disjuncts, but branches only **−13.5%**, **0 classes un-stalled**. `recorded`/`pruned` match the recurrence probe (1017/158891) almost exactly — so the probe was *not* inflated; pruning fires massively. The recurring clashes are **leaves** → each prune saves ~1 branch, never a subtree, because a full clash-dep-set nogood is only satisfied at the leaf.

## What would actually work
**1-UIP asserting clauses** (CDCL: resolution over an implication graph → short clauses that prune subtrees high in the tree). Substantially larger than dep-set recording; the precise next step if the wine-34 / hard-SROIQ-disjunction frontier is pursued.

See `docs/conflict-learning-inc1-results-2026-06-06.md`.

### DIFF
--- a/crates/owl-dl-reasoner/src/lib.rs
+++ b/crates/owl-dl-reasoner/src/lib.rs
@@ -190,6 +190,9 @@ pub fn hyper_sat_probe<A: horned_owl::model::ForIRI>(
     let mut results = Vec::with_capacity(internal.vocabulary.num_classes());
     for (class_id, iri) in internal.vocabulary.classes() {
         let mut engine = HyperEngine::new(&clauses, class_id);
+        if hyper_learning_enabled() {
+            engine = engine.with_learning();
+        }
         let deadline = per_class_timeout.map(|t| std::time::Instant::now() + t);
         let start = std::time::Instant::now();
         let result = engine.decide_with_deadline(max_depth, deadline);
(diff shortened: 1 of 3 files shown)


micheldumontier:
Closing — this was the draft "DO NOT MERGE" Inc 1 spike; the conflict-learning work has since landed differently on `main`. Superseded.

PR #69, opened 2026-08-22
docs: 2026-08-21 handoff — DKey decisions, corpus differential, instrumentation findings

micheldumontier:
**Stacked on #67** so it shows only its own commits. Retarget to `main` once #67 merges.

Documentation only — no code. This is the record the two fix PRs rest on.

## Contents

`docs/handoff-2026-08-21/` — the session handoff: `RESUME-HERE.md` as the entry point, plus `dkey-fix-report.md`, `probe-fix-was-wrong.md`, `rebase-onto-main.md`.

`docs/2026-08-22-dkey-reportedclasses-differential.md` — the corpus differential in full, with raw data in `docs/benchmarks/data-2026-08-22-dkey-*.tsv`.

## Three owner decisions, resolved and recorded

1. **Corpus re-validation — PASS.** Run as a two-arm differential rather than against the Konclude oracle, since `main` is already oracle-validated and the question that discriminates is whether the refactor changed an answer relative to `main`. 663/663 bearing ontologies, 0 DIFFER / 0 REGRESSION / 0 RECOVERY, measured twice (old merge-base and v0.4.22).
2. **`unsatisfiable_bitset()` deletion stands, no code change.** The premise was false: the crate is published only through 0.3.0 while the workspace is at 0.4.22, so the "consumer pinned at 0.4.x" cannot exist. Also records that a rename would *not* have been safer — it removes the old symbol identically while keeping the footgun.
3. **`ReportIdx` newtype: do it, on a follow-up branch after #68 merges.** Internal-only, no semver impact; kept off the branch so it cannot invalidate that branch's evidence.

## Corrections to the earlier record

Several claims in the original handoff were wrong and are retracted in place rather than quietly edited:

- The "false positive live on `main` at default settings" claim, asserted in `rebase-onto-main.md` twice and in a commit message. `main` is sound at that site; the FP exists only *inside* the DKey branch, between its producer conversion and its consumer conversion. The instruction to write a release note for it is withdrawn.
- The open question "did that branch change one end or both?" — answered: both, so it carries no regression.
- "main's lead is docs-only", used to justify not rebasing. It expired within the hour when v0.4.22 landed. Re-check that at the moment of measurement, not before.

## Four instrumentation findings

Each cost real time and each generalises:

1. **`--global-timeout-ms` does not bound the wall** — a known unfixed defect. A sweep wedged ~3 hours against a 60 s budget on `ore_ont_15687` (~180×, vs the 2.9× the existing doc records). Use an external `timeout`.
2. **`sha256("") = e3b0c44298fc` is a false-IDENTICAL generator.** A run exiting non-zero with empty stdout hashes to it in *both* arms and reads as agreement. Assert stdout is non-empty before hashing.
3. **A broken harness fails identically in both arms**, which also reads as agreement — a path typo made every row `BOTH_ERR`. Assert the binaries are executable before sweeping.
4. **Verify stability *within* one arm before comparing two arms.** All 8 NONDET rows this round were contention, not differences; under a 2-run harness every one would have been a confident false finding.

Also noted: the handoff's `RUSTUP_TOOLCHAIN=stable` instruction overrides `rust-toolchain.toml`'s pinned `1.95.0`, which exists precisely to stop clippy drift. It agrees with CI for the errors seen here, but it is the wrong default.

https://claude.ai/code/session_01EWtgfyEhjW5vmzx29kavwu


micheldumontier:
Retargeted from `fix/classify-consistency-probe-aliasing` to `main`. The clippy failure was **inherited, not introduced**: this PR's base predated main's `TBox` backtick fix (`3e9d264`), so the merge under test still contained those two `doc_markdown` errors. #67 was green throughout because it merges into `main`, which has the fix. Against `main` this PR carries #67's commit until that merges, after which the diff reduces to the docs alone.

Issue #35, opened 2026-07-21
materialize_inferred_class_assertions (realization) hangs (>300s) on a 4-axiom ontology; classify + property materialization terminate


rcelebi:
## Summary

`materialize_inferred_class_assertions(...)` (ABox realization) **does not terminate** on a tiny 4-axiom ontology. It was still running after **300 s** (hard cap) on **rustdl 0.3.29**. `classify(...)` and `materialize_inferred_property_assertions(...)` on the same ontology both return immediately — the hang is confined to class-assertion realization.

## Minimal reproducer (1-minimal — dropping any single axiom terminates)

`rustdl_realize_hang.ofn`:

```
Prefix(:=<http://example.org/hang#>)
Ontology(<http://example.org/hang>
  Declaration(Class(:Person)) Declaration(Class(:Man)) Declaration(Class(:Woman))
  Declaration(Class(:Male)) Declaration(Class(:Female))
  Declaration(ObjectProperty(:hasSex)) Declaration(ObjectProperty(:hasParent))
  Declaration(ObjectProperty(:isMotherOf))
  Declaration(NamedIndividual(:a)) Declaration(NamedIndividual(:b))

  EquivalentClasses(:Man   ObjectIntersectionOf(:Person ObjectSomeValuesFrom(:hasSex :Male)))
  EquivalentClasses(:Woman ObjectIntersectionOf(:Person ObjectSomeValuesFrom(:hasSex :Female)))
  ObjectPropertyDomain(:hasParent :Person)
  ObjectPropertyAssertion(:isMotherOf :a :b)
)
```

```python
import rustdl
p = "rustdl_realize_hang.ofn"
rustdl.classify(p)                                   # ✅ returns instantly
rustdl.materialize_inferred_property_assertions(p)   # ✅ returns instantly
rustdl.materialize_inferred_class_assertions(p)      # ❌ never returns (>300 s)
```

## Behaviour matrix (rustdl 0.3.29)

| Operation | Result |
|---|---|
| `classify` | ✅ terminates |
| `materialize_inferred_property_assertions` | ✅ terminates |
| `materialize_inferred_class_assertions` | ❌ **hangs, >300 s, non-terminating** |

## 1-minimality

Removing **any one** of the four non-declaration axioms makes realization terminate:

| Axiom removed | realization |
|---|---|
| `EquivalentClasses(:Man …∃hasSex.Male)` | ✅ terminates |
[...]

PR #71, opened 2026-08-25
fix(absorb): absorb `∃R.{a}` antecedents — #70 non-termination

### LINKED ISSUE #70: instances-expr with ObjectHasValue does not terminate (2-axiom repro; every sibling class-expression form is instant)
## Summary

`instances-expr` does not terminate when the class expression is an `ObjectHasValue` (`p value b`). Every other class-expression form I tried answers the same tiny ontology in 0 seconds. The smallest reproducer is **two axioms**.

The equivalent library call is `class_expression_instances`, so this is not CLI-specific.

## Minimal reproduction

`repro.ofn`, two axioms:

```
Ontology(<http://example.org/o>
    Declaration(ObjectProperty(<http://example.org/p>))
    ObjectPropertyAssertion(<http://example.org/p> <http://example.org/a> <http://example.org/b>)
)
```

```sh
rustdl instances-expr repro.ofn '(<http://example.org/p> value <http://example.org/b>)'
```

**Expected:** `http://example.org/a` (it is asserted directly).
**Actual:** does not terminate. Killed at 30s; on the larger case below, 300s wall with 299s user CPU, so it is spinning rather than blocked.

## It is specific to `ObjectHasValue`

Same ontology (plus `ex:C rdfs:subClassOf ex:D`, `ex:a a ex:C`), release build, 25s cap:

| Query | Result |
| --- | --- |
| `<...D>` (named class) | ok, 0s, returns `ex:a` |
| `(<...C> and <...D>)` (conjunction) | ok, 0s, returns `ex:a` |
| `(<...p> some owl:Thing)` (existential) | ok, 0s, returns `ex:a` |
| `(<...p> value <...b>)` (**ObjectHasValue**) | **hangs** |

Transitivity and subproperties are not involved: I found this on a case with both, then minimised them away and it still hangs. A single non-transitive property with one assertion is enough.

## There is a fast path to the same answer

On SULO plus three triples (`a isPartOf b`, `b isPartOf c`), asking whether `a isIn c` holds (`isIn` is a transitive superproperty of the transitive `isPartOf`):

- `property-values`: **0.00s**, returns `a isIn c` correctly.
- `instances-expr '(<.../isIn> value <.../c>)'`: killed at **300s** (299s user CPU).

So the reasoner can answer it instantly by another route.

## Version

- `rustdl 0.4.23` (`dc662a3`), release build, built from source.
- macOS 26.5.2, arm64.

Also reproduces at `v0.4.22` via the library (found there first).

## Context

Found while building a regression harness for the SULO ontology that uses `owl-dl-reasoner` in-process. The consequence was that the harness had to retire the whole complex-class-expression API family (`class_expression_instances`, `class_expression_entailed_subclass`) and express every query as a bounded `is_class_satisfiable_with_timeout` probe instead, because these calls have no deadline parameter and cannot be interrupted.

Possibly relevant: #48 added complex (anonymous) class expression queries, and `ObjectHasValue` appears to be the one form of that feature which does not terminate.

micheldumontier:
Closes #70.

## The bug

`EquivalentClasses(Q, ObjectHasValue(p, b))` plus one `ObjectPropertyAssertion` — a **two-axiom ontology** — ran for hours on `instances-expr`, `sat-expr` and plain `sat`.

It is not a class-expression bug. Writing the probe axiom by hand and calling `rustdl sat` hangs identically, and in `class_expression_instances` the hanging call is the **completeness-signal probe** (`class_expr_query.rs:220`), not the answer — `instances_of` on the next line returns `a` instantly.

Two layers:

1. **`∃p.{b} ⊑ Q` would not absorb**, leaving the residual `⊤ ⊑ ∀p.¬{b} ⊔ Q` applied at every node. Its `Q` disjunct is *generating*: picking it forces `∃p.{b}`, whose fresh witness is itself nominal, gets the same residual, picks `Q`, generates again. (`tbox-stats`: `residual_gcis: 1`, classified `qualified_exists_antecedent` by `residual-absorbability`.)
2. **`is_blocked_anywhere` refuses to block a nominal node** (`owl-dl-tableau/src/lib.rs:1249`) while `is_blocked_ancestor` has no such exclusion — and `decide` enables anywhere-blocking exactly on the deadline-FREE paths, which is what `is_class_satisfiable` is. So nothing ever cut the cycle.

| arm | `is_blocked_calls` | `is_blocked_true` | outcome |
| --- | --- | --- | --- |
| default (anywhere) | 127,708 | **0** | unbounded |
| `RUSTDL_ANYWHERE_BLOCKING=0` (ancestor) | 480 | **72** | `sat`, instant |
| same residual, **non-nominal** filler, default | 588 | **126** | `sat`, instant |

The third row is the discriminator: identical generating shape, only the nominality of the witness differs. This makes it a regression from the issue #35 v3 fix, which enabled anywhere-blocking *because* ancestor-blocking could not cut a nominal-anchored cycle. On this cycle shape the relation inverts.

## The fix

`RUSTDL_NOMINAL_EXISTS_ABSORPTION` (default ON, `=0` reverts). `∃R.{a} ⊑ ψ` ≡ `{a} ⊑ ∀R⁻.ψ`, so a residual carrying the NNF disjunct `∀R.¬{a}` (or `≤0 R.{a}`) becomes `NominalRule { individual: a, conclusion: ∀R⁻.ψ }` — the nominal-keyed sibling of `absorb_domain_residuals`, and the same move that closed issue #35 v4. A **logical identity**, so sound and completeness-preserving by construction; a differing closure would be a bug, not a trade-off.

Scope is narrower than it looks: `clausify_with_stats` builds from `InternalOntology` axioms, **not** the absorbed TBox (the `clause.rs` module doc saying otherwise is stale), so the wedge cannot be affected — only the main tableau's rule set.

## Evidence, with its limits stated

The pass is **inert on all 7 curated fixtures**, so a green curated closure-diff demonstrates inertness, not correctness — the same trap `datatype_value_membership.rs` documents for DKey. What carries it instead:

1. **Gate scan** over the 448 of 1,920 ORE ontologies containing `ObjectHasValue`/`ObjectOneOf`. A genuine superset: `pool.nominal()` is reachable only from those two lowerings, so no firing ontology can be outside it. **16 FIRES / 430 INERT / 2 pre-existing failures** (`ore_ont_10860` the known SWRL `BuiltInAtom` grammar gap, `ore_ont_8445` a known conversion timeout — identical exit codes in both arms).
2. **Answer identity on the 16 that fire**: `classify --json`, single-thread, 300 s cap — **16/16 byte-identical ON vs OFF**, walls flat, including `ore_ont_7775` where residuals go **253 → 0**.
3. **Revert identity**: flag-OFF is **16/16 byte-identical to the pre-change binary**, so `=0` is a true revert.
4. **FP=0 net**, every fixture present on this host exact: galen 28007, ore-10080-deadline 32356, sio 8904, wine 653, ro 158, sulo 51 — all FP=0 / MISSED=0. Seven REQUIRED fixtures are absent locally (`alehif`, `bibtex`, `pizza`, `notgalen`, `ore-10908-sroiq`, `ore-15672-shoin`) and the script hard-fails rather than skipping, so that run is **partial by provisioning, not by diff**. (`galen` reads 28007 against the 27997 in `CLAUDE.md`; the *oracle* reads 28007 too and predates this work by six weeks, so that is fixture-vintage drift in the design record. `galen` has zero residual GCIs under both arms — a provable no-op — and its output is byte-identical ON vs OFF.)
5. **Canaries**: 6 unit + 5 integration, the integration ones bounded on their own thread so non-termination is an assertable property rather than a CI timeout with no attribution. Two are FP guards: a wrong target individual must entail nothing, and propagation must run to the R-**predecessor** (what `role.flip()` encodes).
6. **Sabotage: 4 run, 4 caught** — disabling the pass, dropping `role.flip()`, accepting `≤k` for `k ≥ 1`, and forgetting the individual each fail exactly the intended canary.

Local: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean, workspace suite **1715 passed / 0 failed / 80 ignored**. (`owl-dl-py` cannot link its test binary on this machine — missing CPython symbols, environmental — so it is excluded from the local run; CI covers it.)

The flag default is pinned behaviourally in `flag_defaults.rs` (table now 31 rows), so changing it is a reviewable edit rather than a silent one.

## Deliberately NOT fixed here

**The blocking asymmetry is a separate, pre-existing defect and remains open.** Absorbing the antecedent removes this *shape*, not the cause underneath it. A nominal `∃`-cycle with **zero** residual GCIs — `Q ⊑ ∃p.{b}` plus `{b} ⊑ Q` — still grows to `RUSTDL_MAX_NODES` and returns *"tableau bailed out without a verdict"* on **both** the pre-change and post-change binaries, while `RUSTDL_ANYWHERE_BLOCKING=0` answers `sat`.

Fixing it means letting `is_blocked_anywhere` fall back to ancestor scope for a nominal `y` (`is_blocked_ancestor` has no nominal exclusion and is already the shipped classify default). That is a blocking-semantics change on every deadline-free query path and needs its own 152-ontology-style bake-off rather than riding on this issue's evidence. It is pinned as the `#[ignore]`d sentinel `residual_free_nominal_cycle_still_exhausts_the_node_cap`, which asserts the **desired** verdict so it trips the moment the gap closes.

Two further items from the report, also unaddressed and worth their own issues: `class_expression_*` take no deadline, and the `incomplete` signal is computed by a probe that can cost unboundedly more than the query it annotates.


micheldumontier:
## FP=0 net now runs nearly in full: **20 passed / 2 failed**, up from 15/7

The gap I flagged in the PR body was mostly a provisioning artefact, and the standing advice for closing it is wrong in a way worth recording: **`fetch-real-ontologies.sh` provisions *inputs* only and never oracles.** `pizza.ofn` was already present — what was missing was `konclude-input/pizza-classified.owx`. Oracles come from `konclude-oracle.sh`, whose loop covers only sio/sulo/ro/wine.

Using a native Konclude v0.7.0-1138 + ROBOT 1.9.10 (no Docker) and the local ORE pool, four of the seven were reconstructible: **pizza** (oracle only), **bibtex** (`ore_ont_3341`), **ore-10908-sroiq**, **ore-15672-shoin**.

### The pipeline was validated before any generated oracle was trusted

The `.kon.owx` artifacts are *not* byte-identical to the committed `-classified.owx` files, so rather than assume equivalence I swapped one in for a fixture whose answer is already known:

```
[fp0] sio: VERIFIED (closure rustdl=8904 oracle=8904 FP=0 MISSED=0)
```

Identical verdict to the committed oracle — the byte differences are serialization, not content. The committed oracle was restored immediately (md5 verified).

### Result

```
[fp0] galen:            VERIFIED (28007 = 28007  FP=0 MISSED=0)
[fp0] galen-5s:         VERIFIED (28007 = 28007  FP=0 MISSED=0)
[fp0] ore-10080:        VERIFIED (32356 = 32356  FP=0 MISSED=0)
[fp0] ore-10908-sroiq:  VERIFIED ( 6001 =  6001  FP=0 MISSED=0)   <- new
[fp0] ore-15672-shoin:  VERIFIED (  142 =   142  FP=0 MISSED=0)   <- new
[fp0] pizza:            VERIFIED (  499 =   499  FP=0 MISSED=0)   <- new
[fp0] bibtex:           VERIFIED (   16 =    16  FP=0 MISSED=0)   <- new
[fp0] sio:              VERIFIED ( 8904 =  8904  FP=0 MISSED=0)
[fp0] wine:             VERIFIED (  653 =   653  FP=0 MISSED=0)
[fp0] ro:               VERIFIED (  158 =   158  FP=0 MISSED=0)
[fp0] sulo:             VERIFIED (   51 =    51  FP=0 MISSED=0)
```

**Four of these independently re-derive `CLAUDE.md`'s own recorded reference values** — pizza 499, bibtex 16, ore-10908 6001, ore-15672 142 — from oracles generated by a *different* reasoner binary than produced the originals, **including the two `corpus.md` attributes to HermiT**. That retires the provenance caveat I raised earlier in this PR: Konclude and HermiT agree on those closures, so the `..._matches_hermit` test names are not a hazard.

This matters for *this* change specifically: `pizza` and `wine` are the corpus's nominal-heavy fixtures, and `ore-10908-sroiq` carries nominals plus inverses and qualified cardinality. A pass that rewrites nominal-filler existentials is now gated against all three.

### Two things that did not close

- **`galen` reads 28007, not the 27997 recorded in `CLAUDE.md`.** The *oracle* reads 28007 too and its file predates this work by six weeks, so it is fixture-vintage drift in the design record rather than a reasoner change — and the four exact matches above are what make that attribution safe rather than convenient. `galen` also has zero residual GCIs under both arms, so this pass is a provable no-op on it.
- **`alehif-test` and `notgalen` remain unobtainable** — no provenance recorded anywhere in `docs/`, and neither is on disk. Konclude's own `Tests/galen-ALEHIF+-classify-request.xml` is *not* the source: it is a 3.1 MB inline-OWLlink galen variant, against `alehif-test`'s 167 classes / 247 closure. Those two hard-fails are the entire residual `2 failed`.

Documented in the second commit so the next person does not repeat the "just run the fetch script" dead end.

micheldumontier:
## Full two-arm regression sweep: 1,920 ORE ontologies, 0 regressions

Pinned binaries — `BEFORE` `7451647e…` built from `main` @ f1ab66b, `AFTER` `6ccfc745…` at branch HEAD. Both arms run per ontology, 60 s cap each, `RAYON_NUM_THREADS=1`. The pre-change binary was rebuilt clean for this: the one used earlier in the PR carried `--features counters`, which compiles in per-rule counter bumps and would have confounded any timing arm.

### Classification

| bucket | count |
| --- | ---: |
| IDENTICAL (byte-identical `classify --json`) | **1,796** |
| BOTH_FAIL (unfinished at 60 s in both arms) | 122 |
| DIFFER | 2 |
| **REGRESSED** (completed before, not after) | **0** |
| **RECOVERED** | **0** |

### Entailments

Across the 1,796 byte-identical ontologies:

| | BEFORE | AFTER | diff |
| --- | ---: | ---: | ---: |
| `direct_subsumptions` | 25,761,815 | 25,761,815 | **0** |
| `unsatisfiable` | 209,269 | 209,269 | **0** |
| `equivalent_groups` | 473,220 | 473,220 | **0** |
| inconsistent KBs | 51 | 51 | **0** |

**25.7 M entailments compared, zero differences.**

### Every flagged row adjudicated — and in all five the pass is INERT

Five rows across the main sweep and the reversed control. Each was adjudicated on two questions: does the pass fire at all (`tbox-stats` ON vs OFF), and is the binary self-consistent across repeats?

| ontology | pass fires? | BEFORE self-consistent? | content-equal? | verdict |
| --- | --- | --- | --- | --- |
| `ore_ont_10517` | no | yes | **yes** | no difference |
| `ore_ont_10807` | no | yes | **yes** | no difference (`incomplete` flag only) |
| `ore_ont_1508` | no | **no** | no | binary disagrees with itself → nondeterminism |
| `ore_ont_9577` | no | yes | **yes** | no difference |
| `ore_ont_16462` | no | yes | **yes** | 60 s cap boundary (55–60 s in every run; AFTER was the *faster* arm in the forward sweep) |

Not one flagged ontology is reachable by the change. Conversely, all 16 where the pass **does** fire were byte-identical in both arms, including `ore_ont_7775` (residuals 253 → 0). `ore_ont_1508` also shows the pass correctly declining a residual of a different shape — the negative controls working on real data.

### Performance: flat — and the raw figure was an artifact

The forward sweep read **−3.90%**, which I did not report, because the ≤1 s bucket showed **−8.10%** and a change that only *adds* a preprocessing pass cannot make small ontologies 8% faster. Arm order was fixed (BEFORE first), so AFTER read a page-cache-warm file. An order-reversed control over 633 ontologies confirms it:

| arm order | apparent AFTER-vs-BEFORE |
| --- | --- |
| forward (BEFORE first) | −3.40% |
| reversed (AFTER first) | **+3.41%** |

The sign flips almost exactly: whichever arm runs **second** is ~3.4% faster. Order-balanced, each arm once per slot — **BEFORE 2699.7 s vs AFTER 2695.2 s = −0.17%**, flat, with **0 ontologies over 1 s more than 1.5× slower**.

### Two method notes now in `CLAUDE.md`

**`incomplete` is a timing flag, not reasoning output, and it breaks byte-comparison of `classify --json`.** Two runs of one unchanged binary on `ore_ont_10807` differ *only* at `"incomplete": false` vs `true`, with all 946 `direct_subsumptions` identical as a set **and in order**, same unsat, same equivalence groups. It reports whether any pair was *cut* by the 5 ms default per-pair deadline, so near that boundary it is a coin flip. Byte-identity still implies content-identity — so every `IDENTICAL` verdict here is sound — but a `DIFFER` is a candidate, not a finding.

**Interleave or balance arm order in any wall comparison.** A fixed order buys a ~3.4% phantom on this corpus.

One correction to an earlier comment of mine in this PR: I initially reported that the BEFORE binary disagreed with itself on `ore_ont_10517` and `ore_ont_10807`. On an idle host it does not — those repeats were competing with the still-running sweep for CPU. The nondeterminism is real but load-dependent, and its clean instance is `ore_ont_1508`; it was never the explanation for those two rows, which are simply unreachable by the change.

PR #75, opened 2026-08-26
fix(datatypes): rdf:langString — a value the reasoner never received (#72)

### LINKED ISSUE #72: rdf:langString: property-values erases the language tag, and DataHasValue never matches one (3-axiom repro)
## Summary

Two problems with `rdf:langString` values, both on a three-axiom ontology:

1. **`property-values` erases the language tag**, so `"bonjour"@fr` and `"bonjour"@de` produce byte-identical output rows. A consumer cannot tell them apart, although under RDF semantics they are distinct literals.
2. **`instances-expr` with a `DataHasValue` over a language-tagged literal returns nothing**, even when the value is asserted directly. The same query shape works for a typed literal.

## Reproduction

`lang.ofn`:

```
Ontology(<http://example.org/o>
    Declaration(DataProperty(<http://example.org/v>))
    DataPropertyAssertion(<http://example.org/v> <http://example.org/de> "bonjour"@de)
    DataPropertyAssertion(<http://example.org/v> <http://example.org/fr> "bonjour"@fr)
)
```

### 1. Tag erased in `property-values`

```sh
rustdl property-values lang.ofn
```

```
http://example.org/de	http://example.org/v	bonjour	http://www.w3.org/1999/02/22-rdf-syntax-ns#langString
http://example.org/fr	http://example.org/v	bonjour	http://www.w3.org/1999/02/22-rdf-syntax-ns#langString
```

The datatype is correctly reported as `rdf:langString`, but the tag itself (`fr`, `de`) is gone, so the two rows differ only in the subject. Any consumer materialising these into a graph produces two triples that are wrong (they claim the same literal) and cannot round-trip.

### 2. `DataHasValue` on a language-tagged literal finds nothing

```sh
rustdl instances-expr lang.ofn '(<http://example.org/v> value "bonjour"@fr)'
```

**Expected:** `http://example.org/fr` (asserted directly).
**Actual:** no instances. Only `warning: verdict is a sound under-approximation (incomplete)`.

Control, same query shape with a typed literal:

```
Ontology(<http://example.org/o>
    Declaration(DataProperty(<http://example.org/v>))
    DataPropertyAssertion(<http://example.org/v> <http://example.org/m> "37.8"^^xsd:double)
)
```

```sh
rustdl instances-expr tmin.ofn '(<http://example.org/v> value "37.8"^^<http://www.w3.org/2001/XMLSchema#double>)'
# -> http://example.org/m     (correct)
```

So the mechanism works; it is specifically the language-tagged case that fails.

Note this terminates, unlike the `ObjectHasValue` case in #70. Different problem, wrong answer rather than no answer.

## Impact

Together these mean a language-tagged value cannot be confirmed by any route: the materialised path loses the tag, and the class-expression path finds nothing. Downstream, this makes `@lang` literals untestable, and makes `property-values` output lossy in a way that is silent.

## Version

`rustdl 0.4.23` (`dc662a3`), release build from source. macOS 26.5.2, arm64. Also observed at `v0.4.22` via the library.

micheldumontier:
Closes #72.

Two reported problems, and a third found while root-causing them.

## 1. `rdf:langString` had no `DKey` bucket, so the literal never reached the reasoner

`data_point_some` tries integer → float → decimal → date → dateTime → `exact_string_literal`. That last one refuses `Literal::Language` — deliberately, per Phase D9. So `data_point_some` returned `None`, the caller raised `ConversionError::UnsupportedDataRange`, and the axiom was **dropped**:

```json
"dropped": { "DataPropertyAssertion: unsupported data range": 2 }
```

The `DataHasValue` probe was not failing to match; there was nothing to match against. This also explains the asymmetry in the report — `property-values` is a **structural** pass that never converts, so it still saw assertions the reasoner never received.

Adds the 8th bucket, `lang:`. **The key is the PAIR `(lexical form, language tag)`**: `"bonjour"@fr` and `"bonjour"@de` are distinct literals, and both are distinct from the `xsd:string` `"bonjour"`. Tags are lowercased at parse (RDF 1.1 §3.3 compares per BCP47, case-insensitively) — a completeness matter, never an FP. Members encode as `hex(lex).hex(tag)` joined by `:`; both halves hex, so neither delimiter can occur inside one and the two-level decode is unambiguous for arbitrary content.

**Why a separate bucket rather than members in `str:`** — the datatypes are disjoint, so sharing would let `"bonjour"` and `"bonjour"@fr` subsume one another. That is the same defect shape as the `xsd:float`/`xsd:double` collision that shipped undetected for months; bucket separation makes it unconstructible rather than merely unlikely.

## 2. `property-values` did not merely erase the tag — it LOST ROWS

`inferred_data_property_values` dropped `lang` and *then* deduped. In the reported fixture the rows differ by subject so both survive; on one subject an assertion disappears:

```
DataPropertyAssertion(v a "bonjour"@de)
DataPropertyAssertion(v a "bonjour"@fr)
DataPropertyAssertion(v a "hello"@en)
```

Three assertions in, **two rows out**. Silent data loss, not lossy formatting.

`quads()` → `quints()`, and `property-values` gains a fifth column. It is emitted **always**, empty for non-langString values, so the row width does not vary by datatype — a variable width would break any consumer splitting on tab, which is the same class of silent lossiness this issue is about. `property_values_text_output` caught the format change, which is what it is for; its expectation is updated with the reasoning recorded in the doc comment.

## 3. New: the dropped-axiom warning reached 3 of 27 CLI commands

`instances-expr` on the fixture printed **nothing at all** — no answer, no warning — while `classify` on the same file reported the drop. `warn_if_dropped` now runs from every command that *answers an entailment question* (14). Pure diagnostics stay out: they report on conversion rather than querying, and `dropped_block` costs an extra `convert_ontology` per call. The warning goes to stderr, so it fires on `--json` paths too, where those commands' JSON carries no `dropped` block; stdout remains a single valid JSON object.

This is `CLAUDE.md`'s own rule — *"When adding a sound under-approximation, ask whether a caller can tell. If not, add the signal in the same change."*

## Evidence — and the corpus is deliberately not part of it

The FP=0 net is **11 fixtures VERIFIED, closures exact** (galen 28007, ore-10908 6001, ore-15672 142, sio 8904, wine 653, pizza 499, ro 158, bibtex 16, sulo 51, ore-10080 32356). But `datatype_value_membership.rs` states plainly that the curated corpus contains no clash of this kind, so a green net here demonstrates **non-regression only**. What actually carries this:

**Konclude ∪ HermiT adjudication, all three reasoners agreeing 4/4:**

| probe | rustdl | Konclude | HermiT |
| --- | --- | --- | --- |
| same lexical + tag, ∈ enumeration | `C ⊑ D` | `SubClassOf(C,D)` | `SubClassOf(C,D)` |
| same lexical, **different tag** | neither | none | none |
| langString vs `xsd:string` | neither | none | none |
| **control**: identical tagged literals | `C ≡ D` | `EquivalentClasses` | `EquivalentClasses` |

The control is load-bearing: both oracles *do* report a relation when one exists, so their silence on the two negative probes is genuine non-entailment rather than the under-reporting Konclude is documented for.

**Tests**: 8 negatives-first canaries in `datatype_value_membership.rs`, 3 unit tests, 3 end-to-end regressions in a new `langstring_values.rs`. Suite **1718 passed / 0 failed**; `fmt` and `clippy --all-targets --all-features -D warnings` clean.

**Sabotage: 4 run, 4 caught** — keying on the lexical form alone, routing langString into the `str:` bucket, restoring the lang-dropping dedup, and skipping tag lowercasing each fail exactly the intended canary.

## One finding worth carrying beyond this PR

`numeric_oneof_parser_matrix_exclusivity` hit `unreachable!()` on the new bucket until fixed — meaning **that half of the FP check was silently not running**. Both parser matrices draw from `sample_iris()` but have separate probes, so extending the fixture list without extending both leaves a decoder pair unchecked. Noted in `CLAUDE.md`.

Also verified rather than assumed: HermiT's output contains `EquivalentClasses(C, g)`-looking entries where `g` is a *data property*. Those are ROBOT echoing the **asserted** axioms, not inferences — my first extraction regex flattened nested axioms and conflated them. The genuine class-to-class inferences are the ones in the table.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1049,6 +1049,16 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   commands to 14** — every command that ANSWERS an entailment question — because
   `instances-expr` printed NOTHING on an ontology whose only two axioms were
   dropped while `classify` on the same file reported them.
+  **AND THE PYTHON STUB WENT STALE WITH CI GREEN.** `data_property_values`'
+  hand-written `__init__.pyi` kept declaring a 4-tuple after the runtime returned
+  5, and BOTH python jobs passed: `test_stubs.py` guards `__all__` NAME drift
+  only, stubs are not checked at runtime, and `test_queries.py` read `q[0]`/`q[1]`
+  so it passed at ANY tuple width. Fixed, plus
+  `test_tuple_returning_stubs_match_runtime_arity`, which parses the declared
+  `list[tuple[...]]` arity out of the stub and compares it against a real call
+  (sabotage-verified: reverting the stub to 4 fails it). **A passing pytest job is
+  NOT evidence the Python surface is correct** — check that the assertion depends
+  on the thing you changed.
 
   **Phase D11 (2026-06-09)** — `DataAllValuesFrom` (unblocked by the D10
   Horn-shortcircuit fix; data-`∀` now routes to the complete hybrid
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1007,6 +1007,49 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   corpus-wide (sio 8904, wine 653, ore-15672 142, ore-10908 6001,
   shoiq-knowledge 449, sulo, ro, bibtex, galen, notgalen, pizza).
 
+  **`rdf:langString` bucket (2026-08-26, #72)** — the 8th DKey bucket, `lang:`,
+  and the one D9 explicitly left out ("language-tagged literals … rejected at
+  parse"). That rejection was a silent DROP, not a sound narrowing you could
+  observe: `data_point_some` fell through every arm to `exact_string_literal`,
+  which refuses `Literal::Language`, so the whole axiom failed conversion.
+  **A language-tagged value was therefore unconfirmable by ANY route** — the
+  reasoning path had no such fact, and `property-values` (a STRUCTURAL pass that
+  never converts) reported it with the tag stripped.
+  **The key is the PAIR `(lexical form, language tag)`**, not the lexical form:
+  `"bonjour"@fr` and `"bonjour"@de` are distinct literals, and both are distinct
+  from the `xsd:string` `"bonjour"`. Tags are LOWERCASED at parse (RDF 1.1 §3.3
+  compares per BCP47, case-insensitively) — a completeness matter, never an FP.
+  Members encode as `hex(lex).hex(tag)`, joined by `:`: BOTH halves hex, so
+  neither delimiter can occur inside one and the two-level decode is unambiguous.
+  **Why a separate bucket rather than members in `str:`** — `rdf:langString` and
+  `xsd:string` are disjoint datatypes (the `Family::LangString` arm already said
+  so), so a shared bucket would let `"bonjour"` and `"bonjour"@fr` subsume one
+  another. That is the SAME defect shape as the v0.4.6–v0.4.9 `xsd:float`/
+  `xsd:double` FP, and bucket separation is what makes it unconstructible rather
+  than merely unlikely. Both parser matrices now cover it (14 buckets);
+  **`numeric_oneof_parser_matrix_exclusivity` was hitting `unreachable!()` on the
+  new bucket until fixed, i.e. that half of the FP check was silently NOT RUNNING** —
+  when extending `sample_iris()`, extend BOTH probes.
+  **Evidence, and the corpus is not part of it.** The FP=0 net is 11 VERIFIED /
+  closures exact, but `datatype_value_membership.rs` says the curated corpus has
+  no clash of this kind, so that shows **non-regression only**. The real gate is
+  8 negatives-first canaries + 3 unit tests + a **Konclude ∪ HermiT adjudication**
+  on 4 probes, all three reasoners agreeing 4/4 — including the DISCRIMINATING
+  CONTROL (identical tagged literals ⇒ both oracles report `EquivalentClasses`),
+  without which their silence on the negative probes would be ambiguous, since
+  Konclude is documented to under-report. **4 sabotages run, 4 caught**: keying on
+  the lexical form alone, routing langString into `str:`, restoring the
+  lang-dropping dedup, and skipping tag lowercasing.
+  Companion fixes in the same change: `inferred_data_property_values` dropped
+  `lang` and THEN deduped, so two tagged assertions on one subject sharing a
+  lexical form **collapsed into one row — losing an assertion, not just a tag**
+  (`quads()` → `quints()`, and `property-values` gains a 5th column, empty for
+  non-langString, kept uniform so consumers splitting on tab do not see a
+  datatype-dependent row width); and `warn_if_dropped` went from **3 of 27 CLI
+  commands to 14** — every command that ANSWERS an entailment question — because
+  `instances-expr` printed NOTHING on an ontology whose only two axioms were
+  dropped while `classify` on the same file reported them.
+
   **Phase D11 (2026-06-09)** — `DataAllValuesFrom` (unblocked by the D10
   Horn-shortcircuit fix; data-`∀` now routes to the complete hybrid
   tableau). Two halves:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1049,6 +1049,28 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   commands to 14** — every command that ANSWERS an entailment question — because
   `instances-expr` printed NOTHING on an ontology whose only two axioms were
   dropped while `classify` on the same file reported them.
+  **FULL TWO-ARM ORE SWEEP (2026-08-26): 1,920 ontologies, 0 regressions, 0 answer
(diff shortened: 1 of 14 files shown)


micheldumontier:
## Correction to this PR's own evidence: the Python stub was stale and CI was green anyway

The first commit widened `data_property_values` from a 4-tuple to a 5-tuple, but the hand-written `__init__.pyi` kept declaring:

```python
def data_property_values(path: str) -> list[tuple[str, str, str, str]]:   # 4
```

Both pytest jobs passed. Three independent reasons they could not see it:

- `test_stubs.py` guards `__all__` **name** drift only, never signatures;
- type stubs are not checked at runtime;
- `test_queries.py` read `q[0]` and `q[1]`, so it passed at **any** tuple width.

So "pytest passes" was not evidence the Python surface was correct — the assertion did not depend on what changed.

Fixed in `3fbc110`:

- Stub corrected, with the tag documented as part of the literal's identity.
- **New `test_tuple_returning_stubs_match_runtime_arity`** — parses the declared `list[tuple[...]]` arity out of the stub and compares it against a real call. Sabotage-verified: reverting the stub to a 4-tuple makes it fail.
- `test_data_property_values` now asserts arity rather than two indices; new `test_data_property_values_keeps_language_tags` covers the row-loss case from Python.
- Three stale doc comments corrected, one of which an earlier edit of mine had left self-contradictory ("must surface as its 4-tuple (subject, property, lexical, datatype, lang)").

Python suite 67 passed / 1 skipped; clippy and fmt clean.

This is the second instance of the same failure mode on this branch — the first being `numeric_oneof_parser_matrix_exclusivity` hitting `unreachable!()`, i.e. half the FP matrix silently not running. Both are now noted in `CLAUDE.md`: **a passing check is not evidence unless its assertion depends on the thing you changed.**

micheldumontier:
## Two-arm ORE sweep: 1,920 ontologies, 0 regressions, 0 answer changes

This PR did not have one; it should have. Adding it, because the risk profile here is **higher** than #71's — that was a logical identity (the same axiom in a different triggered form), whereas this puts axioms into the KB that previously weren't there at all, and adds a disjointness bucket, which is the FP-facing direction.

Pinned binaries (`BEFORE` `7451647e…` from `main`, `AFTER` `523f0e80…`), both arms per ontology, 60 s cap, single-thread, **arm order alternated by index** so the ~3.4% page-cache phantom the #71 sweep hit cancels by construction rather than needing a second reversed run.

| bucket | count |
| --- | ---: |
| IDENTICAL | **1,764** |
| BOTH_FAIL | 123 |
| DIFFER | 33 |
| **REGRESSED** | **0** |
| **RECOVERED** | **0** |

**Entailments over the 1,764 byte-identical ontologies:**

| | BEFORE | AFTER | diff |
| --- | ---: | ---: | ---: |
| `direct_subsumptions` | 24,922,547 | 24,922,547 | **0** |
| `unsatisfiable` | 204,090 | 204,090 | **0** |
| `equivalent_groups` | 407,829 | 407,829 | **0** |

### All 33 DIFFERs adjudicated

Byte-comparison of `classify --json` folds together three things — entailment content, the timing-dependent `incomplete` flag, and the `dropped` reporting block. This change is *expected* to shrink `dropped`, so `DIFFER` is not a defect signal here. Splitting them:

- **27 differ only in the `dropped` block** — the intended effect: a langString axiom that used to be discarded now converts. Entailments identical in every one.
- **5 do not reproduce** on re-run.
- **1 is noise** — see below.
- **0 entailments lost, 0 gained.**

The direction rule applied throughout: a KB that gained axioms can only entail *more*, so gains are expected and losses are the red flag — checked against `unsatisfiable` and `equivalent_groups` too, since proving a class unsat **elides** its Hasse rows and proving classes equivalent **collapses** them, which can make a more-correct answer look like a loss.

### The headline is that nothing moved

langString axioms now enter the KB on **27 real ORE ontologies** and **no answer changes**, because those values participate in no subsumption there. That is exactly why the corpus could never have validated this fix, and why the Konclude ∪ HermiT adjudication in the PR body is the evidence that carries it. The inertness `datatype_value_membership.rs` warns about, demonstrated at corpus scale.

### Wall: flat

Aggregate −1.81%, but the order split brackets it — AFTER-led **+0.88%**, BEFORE-led **−3.60%**. The positional term still dominates, so the sign is not meaningful. One row over 1 s appeared >1.5× slower (`ore_ont_2513`); it has **zero** language literals and on an idle host three interleaved runs per arm are indistinguishable (BEFORE 0.85/0.82/0.80 s, AFTER 0.83/0.78/0.80 s, 165 rows every time), so the sweep figure was host-load noise — and that pair happened to run AFTER first, so it paid the cold-cache cost.

### Method note worth carrying: two runs cannot establish stability

`ore_ont_1508` flagged as **losing** entailments. A two-run same-binary control reported "self-consistent, byte-identical" — and I said so. **The third run refuted it**: BEFORE gives 13928, 13928, **13926**, while AFTER held 13928 across three, which is BEFORE's own modal value. So the "loss" was one sample of each landing on opposite sides of the ontology's noise band.

What makes that reading safe rather than merely convenient is independent of the sampling: `ore_ont_1508` contains **zero** language literals, so `exact_lang_literal` and `parse_lang_range` can never fire, `lang_dkeys` is empty, and both seeding calls iterate an empty vector. There is no code path that can differ.

Recorded in `CLAUDE.md` (`973772f`).

Issue #86, opened 2026-08-30
classify/consistent miss an inconsistency: numeric DKey buckets are non-comparable but not disjoint


micheldumontier:
## Two-axiom reproducer

```
Prefix(:=<http://ex#>)
Prefix(xsd:=<http://www.w3.org/2001/XMLSchema#>)
Ontology(
Declaration(DataProperty(:p)) Declaration(NamedIndividual(:a))
DataPropertyRange(:p xsd:double)
DataPropertyAssertion(:p :a "1.0"^^xsd:float)
)
```

`xsd:float` and `xsd:double` have disjoint value spaces in OWL 2, so the asserted value cannot lie in the declared range and the KB is **inconsistent**.

| | float/double | float/float (control) |
|---|---|---|
| HermiT | `owl:Thing is not satisfiable` | satisfiable |
| Konclude | `false` | `true` |
| **rustdl v0.4.24** | **`consistent`** ✗ | `consistent` ✓ |

Both `rustdl consistent` and `classify --json` report consistent, and `dropped` is empty — so this is not the graceful-degradation path. The float/float control is what makes it a finding rather than a disagreement.

## Root cause

`seed_dkey_subsumptions` buckets DKeys by datatype and seeds subsumption edges **only within a bucket**. That is correct and deliberate — it is what fixed the v0.4.6–v0.4.9 FP where float and double were folded into one f64 bucket and reported *equivalent*.

D11b likewise seeds `DisjointClasses(DKey, DKey)` **only within a bucket**. So two DKeys in different buckets are neither subsumption-related nor disjoint, and `∃p.DKey(f:1.0)` never clashes with an `xsd:double` range.

**The v0.4.9 fix removed the false equivalence and never added the disjointness.** This is its dual, latent since.

## Scope — cross-family already works, only the numeric split is missing

| range | asserted | Konclude | rustdl | |
|---|---|---|---|---|
| `double` | `float` | inconsistent | consistent | **MISS** |
| `decimal` | `float` | inconsistent | consistent | **MISS** |
| `integer` | `float` | inconsistent | consistent | **MISS** |
| `float` | `double` | inconsistent | consistent | **MISS** |
| `integer` | `decimal` | **consistent** | consistent | agree |
| `decimal` | `integer` | **consistent** | consistent | agree |
| `string` | `integer` | inconsistent | **inconsistent** | agree |

## The trap any fix must avoid

[...]

PR #88, opened 2026-08-30
fix(convert): seed DKey disjointness ACROSS datatype buckets (#86)

micheldumontier:
Closes the missed inconsistency root-caused in #86.

## The defect, in two axioms

```
DataPropertyRange(:p xsd:double)
DataPropertyAssertion(:p :a "1.0"^^xsd:float)
```

`xsd:float` and `xsd:double` have disjoint value spaces, so this is **inconsistent**. rustdl reported `consistent`, with nothing dropped.

| | float/double | float/float (control) |
|---|---|---|
| HermiT | `owl:Thing is not satisfiable` | satisfiable |
| Konclude | `false` | `true` |
| rustdl before | **`consistent`** ✗ | `consistent` ✓ |
| rustdl after | `inconsistent` ✓ | `consistent` ✓ |

## Root cause — the dual of a fix already in this codebase

`seed_dkey_subsumptions` buckets DKeys by datatype and seeds subsumption **only within a bucket**. That is correct, and is exactly what fixed the v0.4.6–v0.4.9 FP where float and double were folded into one f64 bucket and reported *equivalent*. D11b seeded `DisjointClasses` only within a bucket too — so cross-bucket DKeys ended up neither comparable **nor disjoint**.

Splitting the buckets removed the false equivalence and never added the disjointness the split implies. `seed_cross_bucket_disjoint` adds it, reusing the same merge-aware component gating so volume stays in the class of the existing seeding.

## Only genuinely disjoint pairs are seeded

**`xsd:integer` is a SUBSET of `xsd:decimal`** — Konclude reports both directions consistent — so that pair is deliberately absent. Seeding it would emit a false `DisjointClasses` and turn this MISS into a false UNSAT, in the subsystem that has already shipped an FP for months. The FP guards are the load-bearing half of the new canary file, not the bug tests.

## Gates

| gate | result |
|---|---|
| canaries | 9/9, incl. 3 FP guards |
| suite | 1856 / 0 |
| clippy, fmt | clean |
| FP=0 net | 20 VERIFIED, every closure exact and unchanged |
| two-arm classify sweep, 637 data-property-bearing ORE | 603 both-ok, **603 identical, 0 DIFFER, 0 `ok→dnf`** |
| two-arm **consistency** sweep, 268 ORE with ≥2 numeric datatypes | **3 changed, all correct; 0 wrong-direction** |
| conversion volume, 3 worst DKey ontologies | identical both arms (`ore_ont_5368` 18,620,251 rules) |

The classify sweep is **non-regression, not benefit** — classify's output is unchanged because the fix lands on the `consistent` surface. The benefit shows in the consistency sweep: the only three verdict changes are `ore_ont_16321`, `ore_ont_4198`, `ore_ont_5014`, all consistent → inconsistent, all Konclude-confirmed.

The FP=0 net's green is **non-regression only** — the curated corpus is inert for DKey work by `datatype_value_membership.rs`'s own admission. The positive evidence is the canaries plus the Konclude ∪ HermiT adjudication.

## Two pre-existing tests asserted the wrong answer

Both adjudicated on their **exact** fixtures before being touched, since a failing test is usually the codebase reporting a defect in the change:

- `float_value_vs_double_range_no_cross_bucket_clash` reasoned from the implementation ("the float assertion simply drops into the float bucket, which has no range constraint") to a semantic conclusion. The half that stands — no cross-**subsumption** — is untouched and preserved in the inverted test's docstring.
- `forall_cross_datatype_no_clash` described itself as a "sound under-approx", which it was; this is the fix for that MISS.

## Known residual, recorded not hidden

This closes the `consistent` / `is_consistent` surface. **`classify` still reports `consistent: true` on `ore_ont_16321`**: the clash is ABox-level, so no *named class* becomes unsatisfiable, and `abox_saturation` has no witness generation to reach the data successor. Closing that needs DKey-aware ABox saturation or a consistency probe on the classify path — different machinery.

`RUSTDL_DKEY_CROSS_BUCKET_DISJOINT`, default ON, `=0` reverts.

Refs #86

### DIFF
--- a/crates/owl-dl-core/src/convert.rs
+++ b/crates/owl-dl-core/src/convert.rs
@@ -3929,8 +3929,22 @@ fn dkey_components(out: &InternalOntology) -> DkeyComponents {
 /// provably never share a label, so skipping them drops zero consumable clash and
 /// keeps the volume in the same class as the within-bucket seeding.
 ///
-/// The caller is responsible for passing only genuinely disjoint bucket pairs —
-/// `xsd:integer`/`xsd:decimal` must NOT be passed here.
+/// The caller is responsible for passing only genuinely disjoint bucket pairs.
+/// Two pairs that look like gaps and are NOT:
+///
+/// * **`xsd:integer` / `xsd:decimal`** — `integer ⊆ decimal`, so they overlap.
+///   `DataPropertyRange(p, xsd:integer)` + `"1.5"^^xsd:decimal` IS inconsistent,
+///   but by VALUE MEMBERSHIP (`1.5 ∉ integer`), not by datatype disjointness;
+///   rustdl misses it and no rule here can close it. Seeding this pair to catch
+///   the `1.5` case would make `"1"^^xsd:integer` in a decimal range a false
+///   UNSAT.
+/// * **`date:` / `dt:`** — rustdl has both buckets, which makes them inviting.
+///   Measured: Konclude reports `date` × `dateTime` CONSISTENT in both
+///   directions, and `HermiT` refuses the input entirely with
+///   `UnsupportedDatatypeException` because `xsd:date` is not in the OWL 2
+///   datatype map (it accepts `dateTime` × `dateTime`, which is what isolates
+///   the cause). No peer supports the entailment, so seeding it would
+///   manufacture a false positive.
 fn seed_cross_bucket_disjoint(
     out: &mut InternalOntology,
     a_keys: &[ClassId],
--- a/crates/owl-dl-core/src/convert.rs
+++ b/crates/owl-dl-core/src/convert.rs
@@ -3109,6 +3109,36 @@ fn seed_dkey_subsumptions(out: &mut InternalOntology) {
     seed_disjoint_bucket(out, &dt_dkeys, OrdRange::disjoint, comp);
     seed_disjoint_bucket(out, &str_dkeys, StrSet::disjoint, comp);
     seed_disjoint_bucket(out, &lang_dkeys, LangSet::disjoint, comp);
+
+    // #86 — CROSS-BUCKET disjointness. Within-bucket disjointness (above) has
+    // always been seeded; cross-bucket never was, so a value of one numeric
+    // datatype asserted against a range of another never clashed:
+    // `DataPropertyRange(p, xsd:double)` + `DataPropertyAssertion(p, a,
+    // "1.0"^^xsd:float)` is INCONSISTENT per Konclude AND HermiT, and rustdl
+    // reported it consistent. That is the dual of the v0.4.6-v0.4.9 FP: folding
+    // float and double into one bucket made them EQUIVALENT, and splitting the
+    // buckets removed the false equivalence without adding the disjointness the
+    // split implies.
+    //
+    // ONLY GENUINELY DISJOINT PAIRS BELONG HERE. `xsd:integer` is a SUBSET of
+    // `xsd:decimal` (Konclude reports both directions CONSISTENT), so that pair
+    // is deliberately absent — seeding it would emit a false `DisjointClasses`
+    // and turn this MISS into a false UNSAT, in the one subsystem that has
+    // already shipped an FP for months. Guarded by
+    // `integer_value_in_decimal_range_is_consistent`.
+    if dkey_cross_bucket_disjoint_enabled() {
+        let ints: Vec<ClassId> = int_dkeys.iter().map(|(c, _)| *c).collect();
+        let floats: Vec<ClassId> = float_dkeys.iter().map(|(c, _)| *c).collect();
+        let doubles: Vec<ClassId> = double_dkeys.iter().map(|(c, _)| *c).collect();
+        let decs: Vec<ClassId> = dec_dkeys.iter().map(|(c, _)| *c).collect();
+        // IEEE floats are disjoint from each other and from the exact numerics.
+        seed_cross_bucket_disjoint(out, &floats, &doubles, comp);
+        seed_cross_bucket_disjoint(out, &floats, &decs, comp);
+        seed_cross_bucket_disjoint(out, &floats, &ints, comp);
+        seed_cross_bucket_disjoint(out, &doubles, &decs, comp);
+        seed_cross_bucket_disjoint(out, &doubles, &ints, comp);
+        // NOT (ints, decs) — xsd:integer ⊂ xsd:decimal.
+    }
     // Numeric-oneof disjointness. FP-CRITICAL DIRECTION: emitting a
     // `DisjointClasses` ADDS clashes, so a wrong "disjoint" is a false UNSAT,
     // not a miss. `BTreeSet::is_disjoint` is exact here because every bucket's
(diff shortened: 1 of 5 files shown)


micheldumontier:
I reached the same defect independently from the other side (root-causing the `ore_ont_16321`/`4198` missed inconsistency) and built a duplicate fix before finding this PR. **I'm dropping mine in favour of this one** — the design here is better in two respects worth naming, and I have two findings that are additive rather than competing.

## Why this PR is the one to take

My version used per-datatype-family marker classes (`DKey ⊑ marker`, markers pairwise disjoint) to get O(#DKeys) instead of pairwise cross-bucket seeding. Two things settle it the other way:

- **Your design can't hit a bug class mine did.** Interning marker classes meant my first cut added 3 classes + 3 axioms to *every* ontology, shifting class ids corpus-wide — walls moved on ontologies with **zero** numeric datatypes (`ore_ont_1016` 0.24 s → 18.77 s). Only a two-arm sweep caught it; my canaries were all green. Seeding pairs directly through the existing component gating interns nothing, so that failure mode doesn't exist. And you verified the volume worry empirically anyway (`ore_ont_5368` identical at 18,620,251 rules).
- **You ran the right sweep.** I ran a classify sweep, which structurally could not see its own targets — `ore_ont_16321` and `4198` came back with *identical* `out_sha256` in both arms, because the fix lands on the `consistent` surface. That's what led me to the classify residual you also document, but it also left me with no positive corpus evidence, which is why mine ships default OFF and yours can justifiably ship ON.

Scope is provably equivalent: your 5 pairs are my 3 family pairs expanded, and my lever independently flips `ore_ont_5014` too — an ontology I'd missed and your consistency sweep found.

## Finding 1 — a residual neither fix can reach (not a bug in this PR)

```
DataPropertyRange(:p xsd:integer)
DataPropertyAssertion(:p :a "1.5"^^xsd:decimal)
```

| | verdict |
|---|---|
| Konclude v0.7.0-1138 | **inconsistent** |
| HermiT 1.4.3 | **inconsistent** |
| rustdl, cross-bucket disjointness ON | consistent |

`xsd:integer ⊆ xsd:decimal`, so they are correctly **not** disjoint and no bucket rule can close this — it needs **value membership** (`1.5 ∉ integer`), which is different machinery. It is also distinct from the classify-surface residual already in your description.

**Your `decimal_value_in_integer_range_is_consistent` canary is correct** — it uses `"1"`, an integral value, where both oracles do say consistent. The direction is value-dependent, not datatype-dependent:

| range | value | oracles |
|---|---|---|
| `xsd:decimal` | `"1"^^xsd:integer` | consistent |
| `xsd:integer` | `"1"^^xsd:decimal` | consistent |
| `xsd:integer` | `"1.5"^^xsd:decimal` | **inconsistent** |

Worth a line in `dkey-numeric-buckets-are-not-disjoint.md` so the pair isn't later read as fully settled, and worth guarding — a future "complete the matrix" change might add `int × dec` on the strength of the 1.5 case and reintroduce the FP this PR is careful to avoid.

## Finding 2 — why omitting `date`/`dateTime` is right, with the reason

Your matrix covers only numerics, which is correct, but the reason is non-obvious and I measured it while scoping mine — worth recording defensively, because `date`/`dateTime` *looks* like a second gap:

```
DataPropertyRange(:p xsd:date)
DataPropertyAssertion(:p :a "2020-01-01T00:00:00"^^xsd:dateTime)
```

- **HermiT: `UnsupportedDatatypeException`** — "supports all and only the datatypes of the OWL 2 datatype map". `xsd:date` is **not in that map**.
- **Konclude: consistent** (both directions, and `date × date` too).

So no peer supports the entailment, and seeding `date ⊥ dateTime` would manufacture a false positive. rustdl has a `date:` and a `dt:` bucket, which makes them an inviting-looking pair for anyone extending `seed_cross_bucket_disjoint` later.

## Incidental: the existing mechanism is coarse-grained

A 7×7 range×value matrix on pre-fix rustdl, which explains why only the numeric interior was broken:

```
range\val  integer decimal double  float   string  date    dateTime
integer     con     con     con     con     INC     INC     INC
decimal     con     con     con     con     INC     INC     INC
double      con     con     con     con     INC     INC     INC
float       con     con     con     con     INC     INC     INC
string      INC     INC     INC     INC     con     INC     INC
date        INC     INC     INC     INC     INC     con     con
dateTime    INC     INC     INC     INC     INC     con     con
```

`{numerics}` / `{string}` / `{temporal}` already clash *across* groups and never *within* — so string-vs-numeric was never the gap, and this PR's numeric-interior scope is exactly the missing piece.

(Caveat on those `date`/`dateTime` `INC` cells: per Finding 2 they are cross-*group* clashes with `string`/numerics, not evidence that `date ⊥ dateTime` should be seeded.)

micheldumontier:
Both findings verified against Konclude and HermiT before recording — neither changes the fix, both are now documented. Pushed in d2a0761.

## Finding 1 — reproduced exactly

| range | value | Konclude | HermiT | rustdl |
|---|---|---|---|---|
| `xsd:integer` | `"1.5"^^xsd:decimal` | inconsistent | inconsistent | **consistent** ✗ |
| `xsd:integer` | `"1"^^xsd:decimal` | consistent | consistent | consistent ✓ |
| `xsd:decimal` | `"1"^^xsd:integer` | consistent | consistent | consistent ✓ |

Value-dependent, not datatype-dependent, exactly as you describe — and unreachable by any bucket rule since `integer ⊆ decimal`.

The guard you asked for already exists implicitly: adding `int × dec` fails **both** int/dec canaries. What was missing was the *reason* at the point of failure, so it now lives in `decimal_value_in_integer_range_is_consistent`'s docstring — where someone "completing the matrix" from the `1.5` case will land — plus a bullet in `seed_cross_bucket_disjoint`'s doc comment, where an extender looks first.

## Finding 2 — reproduced, and the discriminator is what settles it

| range | value | Konclude | HermiT |
|---|---|---|---|
| `date` × `dateTime` | | consistent | `UnsupportedDatatypeException` |
| `dateTime` × `date` | | consistent | `UnsupportedDatatypeException` |
| `date` × `date` | | consistent | `UnsupportedDatatypeException` |
| **`dateTime` × `dateTime`** | | consistent | **is satisfiable** |

I added that last row. It's what makes the finding conclusive rather than suggestive: HermiT accepts `dateTime`, so the exception isolates to `xsd:date` being absent from the OWL 2 datatype map, not to the shape of the probe. Recorded with the same reasoning.

## On the design comparison

The marker-class bug you hit is worth more than the scoping conclusion. Interning classes shifted ids corpus-wide and moved walls on ontologies with **zero** numeric datatypes — green canaries throughout, caught only by a two-arm sweep. That's a sharper instance of the lesson already in this repo's record (curated fixtures are inert for DKey work, so a green net shows non-regression only), and it argues for the pairwise approach on a ground I hadn't considered: not volume, but *blast radius*.

Your read on the sweeps matches what I found from the other side — my classify sweep over 637 ontologies came back 603/603 byte-identical, i.e. structurally blind to its own targets, which is what sent me to the consistency sweep where the three verdict changes actually show. Same trap, same resolution.

`ore_ont_5014`: your lever finding it independently is a useful cross-check — it came out of the 268-ontology consistency sweep here, and it's one of the 7 ontologies the corpus census flagged.

Nothing in either finding needs a code change, so #88 is unchanged apart from the two doc additions. Canaries 9/9, clippy and fmt clean.

micheldumontier:
Investigated your `ore_ont_1016` regression, because if a class-id shift alone could cost 78x that would be a latent fragility firing on *any* future change that interns classes. **It isn't the id shift.** Measured on `ore_ont_1016` — 3,087 classes, zero numeric datatypes:

| variant | wall | mode |
|---|---|---|
| baseline | 0.12 s | `pure EL (saturation-only)` |
| +3 classes **prepended** (shifts every existing id) | 0.11 s | `pure EL` |
| +3 classes appended | 0.11 s | `pure EL` |
| **+3 classes AND 3 `DisjointClasses`** | **1.04 s** | **`hybrid`** |

Prepending shifts every user class id and costs nothing. The **`DisjointClasses` axioms** are what bite, and the route is `saturator_complete_fragment`'s `disjoint_ok` gate: disjointness is admitted only when no functional or inverse-functional role is present, because the disjoint-x-functional-merge interaction is unproven.

`ore_ont_1016` is galen-derived — **150 `FunctionalObjectProperty`, 0 `DisjointClasses`** — exactly the "functional, no disjoint" profile that gate's own comment names as staying on the fast path. Your markers gave it its first disjointness axiom, `has_cardinality_role` was already true, so `disjoint_ok` flipped and the ontology fell to the hybrid path.

So it is a **deliberate conservative fallback, not a defect** — better news than the id-shift theory, since it means nothing about interning is fragile.

The transferable constraint is worth more than either fix:

> An **unconditional** `DisjointClasses` injection knocks every functional-role-bearing pure-EL ontology off the saturation fast path. That population is large (the galen family is well represented in ORE), and the cost is invisible to canaries — it lands as a wall change on ontologies the feature never touches. Only a two-arm sweep sees it, which is exactly how you found it.

It also sharpens why the pairwise design is safer, on firmer ground than "interns nothing": `seed_cross_bucket_disjoint` emits only between actual DKeys inside merge-inducing role components, so an ontology with no numeric DKeys receives **no axioms at all** and its mode is unchanged. Verified rather than assumed — `ore_ont_1016` still reports `# mode: pure EL` with this PR applied, and the 637-ontology two-arm classify sweep found 0 differences.

Recorded in `dkey-numeric-buckets-are-not-disjoint.md` (a3eb788) as a design constraint for anyone adding disjointness anywhere, not just here.

Your magnitude was 78x against my 8.7x — different hardware and a different baseline (0.24 s vs 0.12 s), same direction and same mechanism.

Issue #92, opened 2026-08-31
truncated_probes_are_reported is host-speed flaky (2 of 6 standalone runs fail)


micheldumontier:
`crates/owl-dl-reasoner/tests/realize_incomplete_signal.rs::truncated_probes_are_reported` fails roughly a third of the time, standalone, on an otherwise idle machine.

Measured — 6 consecutive standalone runs, no load, `--release`:

```
run1 ok    run2 ok    run3 FAILED    run4 FAILED    run5 ok    run6 ok
```

## Why

It sets `RUSTDL_REALIZE_PAIR_TIMEOUT_MS=1` and asserts `r.incomplete()` — "a 1 ms per-pair budget must cut at least one case-analysis probe". Whether a probe is cut at 1 ms is host-speed dependent: on a fast machine some runs finish every probe inside the budget, nothing is truncated, and the assertion fires.

This is the same class as the prep-deadline canary that CLAUDE.md records costing PRs #61 and #62 — "chasing the same host-speed canary flake twice".

## Why it matters more than an ordinary flake

It cost me a wrong attribution on this exact tree. A workspace suite run failed here with a new test file present and passed without it, which I read as "my change caused it" on one run per arm. It is not — a second run with the file present was 169 suites / 1862 passed / 0 failed. A flaky gate does not just annoy; it manufactures false causal stories about unrelated changes.

## Possible fixes

- Assert the *observable* rather than the timing: drive truncation deterministically (e.g. a fixture that provably cannot conclude) instead of relying on 1 ms being too short.
- Or make the budget adaptive to a measured per-probe cost, as `RUSTDL_CLASSIFY_INCONSISTENCY_MS` already does for the ABox pre-check.
- Or, weakest, retry-with-smaller-budget before failing.

The sibling test in the same file (`nothing_cut_at_the_default_budget`, asserting `!incomplete()` at the 750 ms default) is the mirror image and presumably flaky in the other direction on a slow or loaded host.

Issue #89, opened 2026-08-31
classify reports consistent:true on a KB that "rustdl consistent" (and Konclude, and KM) call inconsistent


micheldumontier:
Follow-up to #86, which #88 closed on the `consistent` surface only.

## The disagreement, on merged `main` (`f843986`)

```
$ rustdl consistent ore_ont_16321.owl
inconsistent

$ rustdl classify --json ore_ont_16321.owl
"consistent": true, "unsatisfiable": []
```

One binary, one ontology, two surfaces, opposite answers. Konclude reports `owl:Thing ≡ owl:Nothing` and Kobayashi-MaRust reports `consistent: false`, so `consistent` is the correct one and `classify` is wrong.

`ore_ont_4198` behaves identically. Both are in the release population.

## Why `classify` misses it

The clash is **ABox-level**: an individual's data value violates a declared range (`DataPropertyRange(p, xsd:double)` against `"1.0"^^xsd:float`). No *named class* becomes unsatisfiable as a result, so:

* `classify_inconsistency_precheck`'s saturator half (`globally_inconsistent` / `top_is_unsat`) sees nothing — `⊤` is not derived unsatisfiable through any named-class route.
* Its `abox_saturation` half propagates over **named individuals only, with no witness generation** by design (that is what makes it terminate). The clash lives at the *data successor* of the individual, which that fixpoint never materialises.

So the cross-bucket `DisjointClasses` axioms #88 seeds are present in the converted ontology and *are* consumed on the tableau path that `is_consistent` uses — but nothing on the classify path consults them.

## Why it matters

This is the #66 signature: `classify` disagreeing with rustdl's own per-query surface. It bit a downstream consumer once already (a migration tool diffing `classify` hierarchies read the absences as real). Here it is worse than a missing subsumption — a consumer reading `"consistent": true` builds every downstream entailment on a KB that has no model, with no signal that anything is wrong.

It is still **sound** in the usual sense: rustdl under-reports inconsistency, never invents it.

## Two plausible directions

1. **DKey-aware ABox saturation** — teach `abox_saturation` to see a data-successor range violation. Narrow, but adds datatype knowledge to a component deliberately kept structural.
[...]

PR #95, opened 2026-09-01
feat(api): is_consistent_with_timeout — a deadline-bearing consistency check

### LINKED ISSUE #74: No deadline-bearing consistency check; is_class_satisfiable_with_timeout is not a safe substitute (skips the ABox pre-check)
## Summary

There is no deadline-bearing consistency check. `classify` and `is_class_satisfiable` both have `_with_timeout` variants; `is_consistent` has none, so a consistency check cannot be bounded and cannot report "gave up".

The obvious substitute, `is_class_satisfiable_with_timeout` on a top concept, is **not** equivalent, so a caller who reaches for it gets a weaker check than they think.

## The asymmetry at v0.4.23

Bounded variants that exist:

- `classify_with_timeout`, `classify_with_budget`, `classify_n2_with_timeout`, `classify_top_down_with_timeout` (`classify.rs`)
- `is_class_satisfiable_with_timeout` (`lib.rs:5673`)

Consistency entry points, none bounded:

- `is_consistent` (`lib.rs:5774`)
- `is_consistent_internal` (`lib.rs:5784`)
- `is_consistent_with_stats` (`lib.rs:5799`) — returns stats, still no deadline

## Why substituting satisfiability is not sound

`is_consistent_internal_full` runs an ABox-saturation pre-check first (`lib.rs:5814`):

```rust
    if abox_saturation_inconsistent(&internal) {
```

`is_class_satisfiable_with_timeout` has no such step. It converts, saturates, checks unsatisfiability of the queried class, then short-circuits (`lib.rs:5683-5689`):

```rust
    let closure = owl_dl_saturation::saturate(&internal);
    if closure.is_unsatisfiable(class_id) {
        return Ok(Some(false));
    }
    if classify::is_pure_el(&internal) {
        return Ok(Some(true));
    }
```

So on a pure-EL ontology it returns "satisfiable" without ever consulting the ABox. An ABox-only clash, which `is_consistent` catches via the pre-check, is missed.

I measured the two agreeing on every fixture I had, including ABox-only clashes, and still concluded the substitution is unsafe: agreement on a sample is not equivalence, and the divergence is structural and visible in the source. Trading an unbounded check for one that can silently miss an inconsistency is the wrong direction, because a missed inconsistency makes every downstream entailment vacuously true.

## Request

An `is_consistent_with_timeout` (or a deadline parameter on `is_consistent_with_stats`) returning `Ok(None)` on expiry, matching the shape of `is_class_satisfiable_with_timeout`.

## Context

Found building a regression harness that runs a consistency gate before every test case. Every other reasoner call in that harness is bounded; the gate is the sole exception and is documented as such, because there is no bounded form to call.

## Version

`rustdl 0.4.23` (`dc662a3`). Line references are from that tag.

micheldumontier:
Adds `is_consistent_with_timeout(&onto, Duration) -> Result<Option<bool>>`, matching the shape `is_class_satisfiable_with_timeout` already has.

| outcome | meaning |
|---|---|
| `Ok(Some(false))` | a clash was **witnessed** (saturation, `abox_check`, wedge, or tableau) |
| `Ok(Some(true))` | consistent at `is_consistent`'s trust level, including the wedge's trusted `Sat` |
| `Ok(None)` | gave up inside the budget |

**The unbounded path already times out internally and reports `true`.** That's a sound under-approximation but an invisible one — the caller cannot distinguish a real verdict from an exhausted budget. Returning `None` is the point.

The `abox_saturation` pre-check is kept and runs first: dropping it under time pressure would *build* the hazard #74 warns about rather than merely fail to disprove it.

## One correction to the issue

#74's central claim is **structural, not demonstrated**, and I've documented it that way rather than repeating it.

The issue argues satisfiability-of-`⊤` is unsafe because it short-circuits on `is_pure_el` before consulting the ABox. That reading of the source is right — but no divergent case exists. The issue's own author reports measuring the two agreeing on every fixture they had, and four further shapes agree too:

- ABox-only disjointness
- functional-role merge clash
- role-chain-induced range clash
- inverse-materialised domain clash

All agree, because `owl_dl_saturation::saturate` derives the clash and the `closure.is_unsatisfiable` short-circuit fires **before** the `is_pure_el` arm.

**My first canary asserted the divergence and passed — for the wrong reason.** It passed through its `Err(_)` arm, because `owl:Thing` isn't declared in the fixture so the substitute couldn't be called at all. With `owl:Thing` declared it returns `Ok(Some(false))` and agrees. A test that accepts an error as evidence of a behavioural difference proves nothing, and I'd have shipped that claim if I hadn't printed the actual return value.

So the justification rests on what *is* demonstrated:

1. No bounded form existed at all.
2. The unbounded one silently reports `true` on internal expiry.
3. The substitute requires `owl:Thing` to be **declared** and errors otherwise — pinned by `the_substitute_needs_owl_thing_declared`.

`the_substitute_agrees_on_every_shape_tried` pins the measured agreement, and tells the next person to *record* a genuine divergence rather than just fix the test if one ever appears.

## Gates

6 canaries; full suite **169 ok / 1866 passed / 0 failed**; clippy and fmt clean.

Closes #74

### DIFF
--- a/crates/owl-dl-reasoner/src/lib.rs
+++ b/crates/owl-dl-reasoner/src/lib.rs
@@ -5859,6 +5859,103 @@ pub fn is_consistent_with_stats<A: ForIRI>(
     is_consistent_internal_full(internal)
 }
 
+/// Deadline-bearing [`is_consistent`] (#74). `Ok(None)` means the budget expired
+/// without a verdict — the shape [`is_class_satisfiable_with_timeout`] already has.
+///
+/// # Why not just use satisfiability of `⊤`
+///
+/// Two reasons, of unequal strength — stated separately because #74 conflates them.
+///
+/// **Demonstrated:** `is_class_satisfiable_with_timeout` requires `owl:Thing` to be
+/// DECLARED in the ontology and returns `Err(UnknownClass)` otherwise, so it cannot
+/// even be called on most inputs. This function has no such precondition.
+///
+/// **Structural, NOT demonstrated:** #74 argues the substitute is unsafe because it
+/// short-circuits on `is_pure_el` before consulting the `ABox`, so an `ABox`-only clash
+/// would be missed. That reading of the source is right, but no divergent case has
+/// been produced — the issue's author measured the two agreeing on every fixture
+/// they had, and four further shapes (ABox-only disjointness, functional-role merge,
+/// role-chain-induced range, inverse-materialised domain) also agree, because
+/// `saturate` derives the clash and the `is_unsatisfiable` short-circuit fires
+/// first. Do not cite this as a known divergence. See
+/// `tests/bounded_consistency.rs::the_substitute_agrees_on_every_shape_tried`.
+///
+/// This function keeps the `abox_saturation` pre-check regardless, since dropping it
+/// under time pressure would create the hazard rather than merely fail to disprove it.
+///
+/// # What the three outcomes mean
+///
+/// * `Ok(Some(false))` — a clash was WITNESSED (saturation, `abox_check`, the
+///   wedge, or the tableau). Sound: an inconsistency is never invented.
+/// * `Ok(Some(true))` — consistent at the same trust level as [`is_consistent`],
+///   i.e. including the wedge's trusted `Sat`.
+/// * `Ok(None)` — gave up inside the budget. Note the UNBOUNDED
+///   [`is_consistent`] reports `true` in the corresponding internal-timeout case,
+///   which is a sound under-approximation but an invisible one; that silence is
+///   what this function exists to remove.
+///
+/// # Errors
+///
+/// See [`ReasonError`].
+pub fn is_consistent_with_timeout<A: ForIRI>(
+    ontology: &SetOntology<A>,
+    deadline: std::time::Duration,
+) -> Result<Option<bool>, ReasonError> {
+    let start = std::time::Instant::now();
+    let internal = convert_ontology(ontology)?;
+    is_consistent_internal_bounded(internal, start + deadline)
+}
+
+/// Internal entry point for [`is_consistent_with_timeout`], taking an absolute
+/// deadline so conversion time is already charged against the caller's budget.
+///
+/// # Errors
+///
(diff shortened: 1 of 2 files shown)


micheldumontier:
Pushed `b21c866` — the API overran its own deadline, found by measuring it while scoping #89.

I'd left the ABox-saturation pre-check unbounded, reasoning it was cheap and that skipping it would rebuild the hazard #74 warns about. It isn't always cheap: **`family.ofn` took 1.17 s against a 500 ms budget — a 2.3x overrun.** A deadline-bearing API that overruns its deadline is a weak contract.

`abox_saturation_inconsistent_bounded` already existed (classify uses it), so the fix passes the caller's *remaining* budget. A timeout there yields `false` ("no clash in budget"), a sound under-approximation, and the deadline check immediately after turns that into `None` rather than a guess. The pre-check still runs first — bounded, not dropped.

Measured on `family.ofn`:

| budget | elapsed | ratio | result |
|---|---|---|---|
| 100 ms | 104 ms | 1.03x | `Ok(None)` |
| 500 ms | 509 ms | 1.02x | `Ok(None)` |
| 2000 ms | 1.0 s | 0.58x | `Ok(Some(false))` |

Honest `None` below the needed budget, a real verdict above it.

## The canary's first version was useless, and only a sabotage revealed it

400 individuals with no transitive role costs milliseconds either way, so restoring the unbounded call left the test **green**. What makes the pre-check expensive is the **role-chain closure** — the same thing that makes `family` slow:

| n | bounded | unbounded |
|---|---|---|
| 200 | 447 µs | 63 ms |
| 400 | 284 µs | 380 ms |
| **800** | **557 µs** | **4 s** |

Re-sized to n=800 with a 2 s threshold — ~3500x above the passing cost and ~2x below the failing one, so it isn't the host-speed knife-edge #92 documents. The sabotage now fails it at 3.59 s with a message naming the measurement, and the table lives in the test so the sizing can be re-derived rather than guessed.

7 canaries, 15/15 stable at default concurrency; full suite **169 ok / 1867 passed / 0 failed**.

Issue #49, opened 2026-07-24
Complete nominal + number-restriction realize/consistency (decide #35 v4, not just bound it)


micheldumontier:
## Goal
Make `realize` / `is_consistent` **decide** the issue #35 v4 nominal + number-restriction pattern (HermiT/Konclude parity), not just terminate with a sound under-approximation.

## Background
v0.3.40 fixed the *hang* on this pattern with a sound safety net (default per-pair `RUSTDL_REALIZE_PAIR_TIMEOUT_MS` + `RUSTDL_MAX_NODES` cap → `Ok(None)` / MISS). It is **not** a completeness fix: on the reproducer, `realize` omits the entailed types HermiT reports.

Reproducer (`realize` returns a sound MISS; HermiT decides it in 0.5s):
```
SubClassOf(:A ObjectOneOf(:x :y :z))
EquivalentClasses(:B ObjectIntersectionOf(:A ObjectMinCardinality(2 :r :C)))
ObjectPropertyDomain(:r :A)
```

## Root cause
The completion graph grows without bound: `ObjectPropertyDomain(r,A)` absorbs to an untriggered residual `⊤ ⊑ ¬∃r.⊤ ⊔ A`; choosing its `A` disjunct on a fresh `≥2 r.C` witness re-opens the covering nominal disjunction, and the o-rule folds the witness back into the constraint owner → endless regeneration. Blocking cannot cut it: every cycle node is nominal-tainted, and both `is_blocked_anywhere`/`is_blocked_ancestor` exclude nominal nodes on both sides (`crates/owl-dl-tableau/src/lib.rs:1021` / `:1062`), because blocking a singleton is unsound in general.

## What was already tried and rejected
**Nominals-first scheduling** (`RUSTDL_NOMINAL_FIRST`, implemented but **default OFF / dormant**): defer `∃`/`≥` generation until the covering `ObjectOneOf` disjunction resolves and the o-rule merges the node into its canonical nominal. **Falsified** — the domain-residual path re-opens generation through a route the concept-rule-keyed guard never intercepts; the graph still diverges (verified: bounded-node gate `nominal_first_bounded.rs::issue35_v4_completion_graph_is_bounded`, `#[ignore]`d). Scheduling is the wrong mechanism. The machinery is left dormant as scaffolding — reuse or remove.

## Candidate approaches
[...]

Issue #81, opened 2026-08-28
Range restrictions are not folded into NESTED existential witnesses: pure-EL entailment and unsatisfiability both MISSED (incomplete:false)


micheldumontier:
Found while designing the negative-certificate verifier; not by running it. Possibly the same root cause as #80 (nested existential markers not re-closed), but the missing information arrives via `ObjectPropertyRange` rather than a told subsumption, so filing separately for triage.

## Reproducer A — entailment missed, rustdl emits ZERO rows

```
Prefix(:=<http://ex.org/>)
Ontology(<http://ex.org/cascade>
ObjectPropertyRange(:r :F)
ObjectPropertyRange(:u :G)
SubClassOf(:A ObjectSomeValuesFrom(:r ObjectSomeValuesFrom(:s :Y)))
SubClassOf(ObjectIntersectionOf(ObjectSomeValuesFrom(:s :Y) :F) ObjectSomeValuesFrom(:u ObjectSomeValuesFrom(:v :W)))
SubClassOf(ObjectIntersectionOf(ObjectSomeValuesFrom(:v :W) :G) :Z)
SubClassOf(ObjectSomeValuesFrom(:u :Z) :P)
SubClassOf(ObjectSomeValuesFrom(:r :P) :FINAL)
)
```
(declarations elided for brevity; full file in fixtures)

`A ⊑ FINAL` is entailed, by a seven-step derivation:

```
A ⊑ ∃r.(∃s.Y)                  -> A has an r-successor w1 ∈ ∃s.Y
Range(r,F)                     -> w1 ∈ F, so w1 ∈ (∃s.Y ⊓ F)
(∃s.Y ⊓ F) ⊑ ∃u.(∃v.W)         -> w1 has a u-successor w2 ∈ ∃v.W
Range(u,G)                     -> w2 ∈ G, so w2 ∈ (∃v.W ⊓ G)
(∃v.W ⊓ G) ⊑ Z                 -> w2 ∈ Z
∃u.Z ⊑ P                       -> w1 ∈ P
∃r.P ⊑ FINAL                   -> A ⊑ FINAL
```

| reasoner | result |
|---|---|
| **rustdl** | `direct_subsumptions: []` — **zero rows**, `incomplete: false` |
| Konclude v0.7.0-1138 | `[(A, FINAL)]` — its only non-trivial row |

Banner: `# fragment: pure-EL (… saturator alone is complete)`.

[...]

Issue #78, opened 2026-08-28
Wedge reports Sat for a subsumed pair on a 6-axiom ALC ontology (deferred=0, no caps hit) — engine cause of #66


micheldumontier:
## Summary

The hypertableau wedge reports **`Sat`** for `Cap ⊓ ¬VE` on a six-axiom ALC ontology where the subsumption genuinely holds. The main tableau refutes the same probe immediately. Because `RUSTDL_HYPERTABLEAU_TRUST_SAT` defaults ON, `classify` believes the wedge and silently drops the subsumption.

This is the engine-level cause of #66. That issue is about the *consequence* (`classify` disagreeing with `subclass`, with no signal); this one is about the wedge being wrong on a pattern it fully claims.

**The evidence says this is a small, localised bug rather than an inherent incompleteness** — see "why this looks tractable" below.

## Reproducer

`min66.ofn` — six logical axioms, ALC, no nominals, no datatypes, no inverses:

```
Prefix(:=<http://ex#>)
Ontology(
Declaration(Class(:A)) Declaration(Class(:V)) Declaration(Class(:R))
Declaration(Class(:Pizza)) Declaration(Class(:Cap)) Declaration(Class(:VE))
Declaration(ObjectProperty(:p))
SubClassOf(:A :V)
SubClassOf(:Cap :Pizza)
SubClassOf(:V :R)
DisjointClasses(:Pizza :R)
EquivalentClasses(:VE ObjectIntersectionOf(:Pizza ObjectAllValuesFrom(:p ObjectUnionOf(:V ObjectComplementOf(:R)))))
SubClassOf(:Cap ObjectAllValuesFrom(:p ObjectUnionOf(:A ObjectComplementOf(:R))))
)
```

`Cap ⊑ VE` holds: `∀p.(A ⊔ ¬R) ⊑ ∀p.(V ⊔ ¬R)` given `A ⊑ V`. HermiT agrees (`robot reason` materialises `SubClassOf(Cap VE)`).

```console
$ rustdl subclass min66.ofn 'http://ex#Cap' 'http://ex#VE'
yes                                   # main tableau — correct

$ rustdl classify min66.ofn | grep Cap
direct  http://ex#Cap  http://ex#Pizza # Cap ⊑ VE MISSING

$ RUSTDL_HYPERTABLEAU=0 rustdl classify min66.ofn | grep -c 'Cap.*VE'
1                                     # wedge disabled -> classify is correct
```

[...]

Issue #76, opened 2026-08-27
Unsound: subclass reports a false subsumption — ≤n merge-pair choice is deterministic (5-axiom repro; live on pizza)


micheldumontier:
## Summary

`rustdl subclass` reports a subsumption that does not hold. **Five axioms**, no datatypes, no nominals, no inverses:

```
Prefix(:=<http://ex.org/p#>)
Ontology(<http://ex.org/p>
SubClassOf(:A ObjectSomeValuesFrom(:r :M))
SubClassOf(:A ObjectSomeValuesFrom(:r :T))
SubClassOf(:A ObjectSomeValuesFrom(:r :PT))
DisjointClasses(:M :PT)
EquivalentClasses(:I ObjectMinCardinality(3 :r))
)
```

```console
$ rustdl subclass r10b.ofn 'http://ex.org/p#A' 'http://ex.org/p#I'
yes          # WRONG
```

**Expected `no`.** Only `M` and `PT` are disjoint, so the `T`-successor may merge with either one, giving a model with two `r`-successors. `≥3 r` is not entailed.

HermiT and Konclude both report no subsumption.

Present on a clean build of `b24f1f9` (current `main`).

## The controls are what make this unambiguous

| fixture | disjointness | correct answer | rustdl | HermiT | Konclude |
| --- | --- | --- | --- | --- | --- |
| `r10b` | `Disjoint(M, PT)` | **no** | **yes** ❌ | no | no |
| `r10c` | `Disjoint(M, T, PT)` (all pairwise) | **yes** | yes ✅ | yes | yes |
| `r10` | `Disjoint(M, T)` | **no** | no ✅ | no | no |

`r10c` is a discriminating positive control: with all three pairwise disjoint the subsumption genuinely holds, and all three reasoners agree. So the oracles are not merely silent on this relation.

`r10` vs `r10b` is the tell: **the verdict flips depending on which pair happens to be disjoint**, while the semantics of "is a third successor forced?" is unchanged. That is order-dependent determinism, not a semantic distinction.

## It is also live on the pizza corpus fixture

```console
$ B=http://www.co-ode.org/ontologies/pizza/pizza.owl
$ rustdl subclass ontologies/real/pizza.ofn "$B#Margherita" "$B#InterestingPizza"
yes          # WRONG, reproducible 3/3, `explain` says "answered by tableau"
```

[...]

PR #98, opened 2026-09-03
fix(classify): do not trust a wedge Sat for cardinality over a complex qualifier

### LINKED ISSUE #91: classify misses an unsatisfiable class when the cardinality qualifier is complex
```
SubClassOf(:A ObjectMaxCardinality(1 :r ObjectIntersectionOf(:B :C)))
SubClassOf(:A ObjectMinCardinality(2 :r ObjectIntersectionOf(:B :C)))
```

`A` is unsatisfiable — it needs ≥2 and ≤1 `r`-successors in the *same* class.

| | classify `unsatisfiable` | per-pair `rustdl sat A` |
|---|---|---|
| complex filler `ObjectIntersectionOf(:B :C)` | **`[]`** — MISSED | `unsat` ✓ |
| atomic filler `:B` (control) | `['A']` ✓ | `unsat` ✓ |

Konclude and HermiT both report `EquivalentClasses(owl:Nothing, A)`. rustdl's own per-pair `sat` is right in both rows; only `classify` is wrong, and only with the complex filler. Stable across 3 runs.

The atomic control is what makes this a **qualifier-shape** defect rather than a cardinality defect.

## Not the trust_sat mechanism

Unlike the sibling finding in #90, **no flag recovers it**:

| config | `unsatisfiable` |
|---|---|
| default | `[]` |
| `RUSTDL_HYPERTABLEAU_TRUST_SAT=0` | `[]` |
| `RUSTDL_CLASSIFY_VERIFY_REFUTATIONS=1` | `[]` |
| `RUSTDL_SPIKE_CARD_ATOM=1` | `[]` |

So it is distinct from the wedge-`Sat`-trusted silent MISS that explains #66, #78 and #90. It lives in whatever `classify` uses to decide class satisfiability, which on this input disagrees with the `sat` subcommand.

Same *family* as #66's secondary observation (classify trusting saturation for satisfiability). That observation was pizza-specific and is now resolved on pizza; this is a live instance of the same disagreement on a different shape.

## Where it was found, and a hazard it retires

Found while probing a *different* recorded suspicion — that `fresh_class_id` scans `Atom::Class` and `Atom::Exists` but not `Atom::AtMost(_, Some(c), _, _)`, so an ontology whose largest class id occurs only in an `AtMost` qualifier could alias real classes.

**That hazard is REFUTED — do not spend time on it.** Every call site clamps `fresh_class_id(&base).index().max(num_classes)`, and real ids are all `< num_classes`; and a qualifier that is not a named class goes through `atomic_name_of`, which emits `Q(x) → filler(x)`, so the synthetic id appears in a `Class` atom and IS scanned. The fixture built to demonstrate that FP produced no FP — it produced this MISS instead.

## Severity — corpus reach measured

Census over all 1,920 ORE ontologies: only **5** carry a qualified cardinality whose filler is a constructor rather than a named class, and **all 5 show zero unsat disagreement** against a Konclude oracle. Pre-existing (identical on the pinned v0.4.23 control), completeness-only, silent.

Low measured reach, but unlike #90 the shape is not exotic — complex cardinality qualifiers are ordinary modelling — so it may have reach outside ORE.

Full analysis: `docs/known-limitations/classify-misses-unsat-complex-cardinality-qualifier.md`.

micheldumontier:
`A ⊑ ≤1 r.(B ⊓ C)` with `A ⊑ ≥2 r.(B ⊓ C)` makes `A` unsatisfiable. Konclude and HermiT agree, `rustdl sat A` agrees — and `classify` reported `unsatisfiable: []`.

## The third instance of one pattern

The previous two sit in the very same `needs_verify` expression:

| set | construct the wedge cannot count |
|---|---|
| `data_counting_classes` | concrete-domain (DKey) cardinality |
| `nominal_counting_classes` (#49) | nominal counting — its own comment already reads *"Same defect one construct over"* |
| **this (#91)** | **object cardinality over a COMPLEX qualifier** |

classify's unsat probe trusts the wedge's `LabelOracle::Sat` unless `needs_verify` fires. `cardinality_qualifier` Tseitin-names a complex filler, so the wedge counts a synthetic name without relating it to the members and can't see that the `≤` and `≥` range over the **same set**. With a *named* filler it can — which is what isolates this to the qualifier shape rather than to cardinality, and why `the_atomic_filler_control_still_works` is evidence rather than decoration.

**Why the issue reported that no flag recovers it:** `TRUST_SAT=0`, `CLASSIFY_VERIFY_REFUTATIONS=1` and `SPIKE_CARD_ATOM=1` all target the *subsumption* path. This is the *satisfiability* probe, which had no escape hatch for this construct.

## Direction of risk

Withdrawing trust only swaps a wedge `Sat` for the complete tableau path, so it cannot invent an unsatisfiable class — being wrong costs wall time, not correctness.

Both negative controls are **oracle-adjudicated**, not assumed: compatible bounds (`≥2` with `≤3`) and *different* complex fillers (`≤1 r.(B⊓C)` with `≥2 r.(B⊓D)`) are `[]` per Konclude **and** HermiT. The second is the shape a careless fix would break by treating any two complex qualifiers on one role as the same set.

## Cost

Default ON, unlike its `RUSTDL_NOMINAL_COUNTING_VERIFY` sibling, because the addressable set is measured and tiny: **5 of 1,920** ORE ontologies author a qualified cardinality with a non-named filler, and the set is empty otherwise so the clause short-circuits. Curated fixtures — identical output, flat wall: pizza, ro, sulo, bibtex, sio, go-basic.

## wine first read as a real output change, and is not

Same-arm repeats differ from **each other** (run 3 of 4 diverged), so wine is nondeterministic within one binary and one flag setting — exactly what the design record says about a truncating per-pair budget. At a non-truncating 2000 ms budget both arms give an identical row-set hash, 197 rows.

Cross-arm single-run comparison on a nondeterministic input isn't evidence. Second time that trap appeared this session.

## The gate fired again

#93's agreement gate reported `the gate did NOT detect ["#91"]` — second consecutive day it has caught a tracked defect being fixed. Its fixture moved into `previously_divergent_shapes_now_agree`.

Worth noting it reported **three** divergences for this one shape: the missed unsat also made `A ⊑ B` and `A ⊑ C` wrong, and all three cleared together — confirming a missed unsat propagates into subsumptions.

## Gates

5 canaries, sabotage-verified (flag OFF fails exactly the detection test and no control); full suite **175 ok / 1879 passed / 0 failed**; clippy and fmt clean.

Closes #91

### DIFF
--- a/crates/owl-dl-reasoner/src/classify.rs
+++ b/crates/owl-dl-reasoner/src/classify.rs
@@ -3801,7 +3801,26 @@ fn classify_top_down_internal_impl(
                                     || closure
                                         .subsumers_of(class_id)
                                         .iter()
-                                        .any(|s| prepared.nominal_counting_classes.contains(s))));
+                                        .any(|s| prepared.nominal_counting_classes.contains(s))))
+                            // THE SAME DEFECT ONE CONSTRUCT FURTHER (issue #91). The
+                            // wedge does no OBJECT cardinality counting over a COMPLEX
+                            // qualifier: `A ⊑ ≤1 r.(B ⊓ C)` with `A ⊑ ≥2 r.(B ⊓ C)`
+                            // makes `A` unsatisfiable, Konclude and HermiT agree, and
+                            // `rustdl sat A` agrees — but `classify` reported
+                            // `unsatisfiable: []`. `cardinality_qualifier` Tseitin-names
+                            // the complex filler, so the wedge counts a synthetic name
+                            // without relating it to the members and cannot see that the
+                            // `≤` and `≥` are over the SAME set. The ATOMIC-filler
+                            // control decides correctly, which is what isolates this to
+                            // the qualifier shape rather than to cardinality.
+                            || (crate::complex_qualifier_verify_enabled()
+                                && !prepared.complex_qualifier_counting_classes.is_empty()
+                                && (prepared
+                                    .complex_qualifier_counting_classes
+                                    .contains(&class_id)
+                                    || closure.subsumers_of(class_id).iter().any(|s| {
+                                        prepared.complex_qualifier_counting_classes.contains(s)
+                                    })));
                         if !needs_verify {
                             return Ok((i, true, true));
                         }
--- a/crates/owl-dl-reasoner/src/lib.rs
+++ b/crates/owl-dl-reasoner/src/lib.rs
@@ -6848,7 +6848,7 @@ fn concept_has_complex_qualifier_counting(pool: &ConceptPool, c: ConceptId) -> b
 /// qualified cardinality whose filler is not a named class, and
 /// `complex_qualifier_counting_classes` is EMPTY for every other input, so the
 /// clause short-circuits before touching the closure.
-fn complex_qualifier_verify_enabled() -> bool {
+pub fn complex_qualifier_verify_enabled() -> bool {
     std::env::var_os("RUSTDL_COMPLEX_QUALIFIER_VERIFY").is_none_or(|v| v != "0")
 }
 
--- a/crates/owl-dl-reasoner/src/lib.rs
+++ b/crates/owl-dl-reasoner/src/lib.rs
@@ -6507,6 +6507,12 @@ pub(crate) struct PreparedOntology {
     /// `build_data_counting_classes`). The classify unsat-probe
     /// main-tableau-verifies these instead of trusting the wedge's `Sat`.
     pub(crate) data_counting_classes: std::collections::HashSet<owl_dl_core::ir::ClassId>,
+    /// Classes carrying a `Min`/`Max` over a COMPLEX qualifier (#91). Consumed by
+    /// classify's unsat probe exactly as `data_counting_classes` is. Empty unless
+    /// the ontology authors a qualified cardinality whose filler is not a named
+    /// class, so the overwhelming majority of inputs pay nothing.
+    pub(crate) complex_qualifier_counting_classes:
+        std::collections::HashSet<owl_dl_core::ir::ClassId>,
 
     /// Classes the wedge cannot decide for lack of nominal counting (see
     /// `build_nominal_counting_classes`). Consumed by classify's unsat probe the same way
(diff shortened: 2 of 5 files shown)


micheldumontier:
## Review: the fix stands; the default-ON flip is now measured. One blocking nit.

I re-derived the evidence rather than reading it, and separated the two questions —
**the fix** stands on its canaries and oracle adjudication; **the default** needed a
corpus arm it didn't have. That arm now exists and says ON is safe.

### The new default-ON flag was not pinned — fixed in ec326e6

`RUSTDL_COMPLEX_QUALIFIER_VERIFY` is absent from `tests/flag_defaults.rs`, while its
closest sibling — `RUSTDL_COUNTING_PAIR_VERIFY`, also default-ON, also a wedge-trust
withdrawal — **is** pinned there (line 186). CLAUDE.md's guardrail is explicit that a
default change requires editing that table in the same commit, and that table exists
*because* `RUSTDL_CLASSIFY_LABELS_AMORTIZE` drifted undetected.

Pushed as the sibling's shape: `complex_qualifier_verify_enabled` becomes `pub` and
the row asserts `true`. Sabotage-verified — flipping the accessor to
`is_some_and(|v| v == "1")`, the silent drift it exists to catch, fails
`public_bool_flag_defaults_are_pinned`; restored, 3/3 green; clippy and fmt clean.

### The "5 of 1,920" is grep-derived, and the gate figure is 4

I reproduced the 5 exactly with a same-line grep, which is what identifies it as a
grep rather than a gate measurement — the distinction Lever 1 already paid for
("~40 recoveries VERIFIED BY GATE-PROBE, not the 67 grep guess"). A
whitespace-insensitive scan finds **7**, adding two `DataExactCardinality → DataOneOf`
ontologies an object-only pattern cannot see.

Measured by gate, via the `# satisfiability probes: tableau=` delta ON vs OFF, the
predicate fires on **4**:

| ontology | tableau probes OFF → ON |
|---|---|
| `ore_ont_11647` | 0 → **20** |
| `ore_ont_15514` | 0 → **4** |
| `ore_ont_9012` | 0 → **1** |
| `ore_ont_9540` | 13 → **15** |

It correctly does **not** fire on the two `DataOneOf` ones — their filler lowers to an
atomic DKey, exactly as the design predicts — nor on `ore_ont_668`. So the count is
right for the wrong reason: two members the grep found don't fire, and two it missed
don't either.

Frame validity is by construction, not by grep: only the three cardinality axioms
construct `Min`/`Max` (`convert.rs:1037/1042/1047`), and NNF **preserves** filler
complexity rather than creating it (`normalize.rs:68/72/123/129` carry `c_nnf`
through), so an ontology with no `(Object|Data)(Min|Max|Exact)Cardinality` cannot
reach the predicate. That set is **497**, and I swept all of it.

### Two-arm sweep, 497-ontology superset frame: 0 answer changes, 0 outcome changes

One pinned binary (`0be4cfb516f6`), the env var the only difference and recorded in
both manifests, 60 s cap, single-thread.

- **452 ok/ok, 44 dnf/dnf, 1 err/err — no `ok → dnf`.**
- **447 answer-triples IDENTICAL, 50 both-empty, 0 DIFFER.**
- Aggregate wall **−0.21%**; **0 ontologies both >1 s slower and >1.5×**.

Compared as the triple (direct rows, `unsat`, equivalence groups) over non-`#` lines,
so every timing banner and `incomplete` is excluded by construction. That mattered
here: this fix's success mode is *finding* an unsatisfiable class, which **elides**
that class's Hasse rows — so a row-count comparison alone would read a correct fix as
a regression. It's the direct-vs-closure trap, and `ore_ont_8475` is on record for it.

### The one real cost worth recording

`ore_ont_11647`: **1.43 s → 1.76 s (+23%)**, identical answers, reproducible across 3
sequential repeats with arm order alternated (ON 1.76/1.75/1.76, OFF 1.44/1.43/1.42).
Attributable — it's the ontology taking 20 extra tableau probes. Small, but it is the
flip's actual price and belongs in the PR body rather than "flat wall".

### `ore_ont_9540` is safe, but do not read a cross-arm hash on it

This is the documented label-cache guard case, so I gave it the most attention.

Arm ON is **self-inconsistent: 7 of 8 runs one row-set, 1 of 8 another** (the
cold-cache first run); arm OFF is 8/8 stable. The deviation is **one Hasse row** —
`UJI_Wall ⊑ Object_type` where the other runs derive the nearer
`UJI_Wall ⊑ Possible_UJI_Wall` — i.e. a run that failed to derive the nearer parent.
A sound truncation miss, the #76/`ore_ont_1508` signature.

1-in-8 vs 0-in-8 is not separable from noise, and "extra probes steal budget from the
tier walk" is a plausible mechanism, so I ran the control instead of arguing:
**at `--pair-timeout-ms 1000` both arms are identical — 41 rows, same hash, 217.34 s
vs 217.22 s.** Budget truncation, not the fix. Same conclusion the PR reached for
`wine`, by the same route.

Wall at the default is flat either way (14.18–14.25 s across all 6 interleaved runs).

### The corpus is inert for this fix, and that is the headline

On all 4 ontologies where the predicate fires, the main tableau **confirms** the
wedge's `Sat` — no new unsatisfiable class anywhere in 497. So ORE cannot validate
this fix; its evidence remains the 5 canaries plus the Konclude/HermiT adjudication,
and the sweep is non-regression only. Same shape this repo has recorded for #70, #72,
#78 and #81 — worth saying explicitly in the PR body so a future reader doesn't
mistake the green sweep for confirmation.

### Verdict

Sound, correctly scoped, symmetric with both siblings (I checked: the nominal sibling
applies the identical `ConceptExpr::Atomic(sub)` restriction, so the narrowness is
consistent and deliberate, not an oversight). Nothing in the wall data argues for
downgrading the default to OFF. **Merging.**

The two items above worth folding into the record when convenient: `ore_ont_11647`'s
+23% as the flip's measured price, and the ORE-inertness note so a future reader does
not mistake this green sweep for confirmation of the fix.

Issue #42, opened 2026-07-24
Datatype (SROIQ(D)) expressivity gaps: nested composites, non-string DataOneOf, data cardinality counting


micheldumontier:
Remaining **datatype (SROIQ(D)) expressivity** drops (all sound under-approximations — the construct is dropped, never mis-decided):

- [x] **Nested composite data ranges** — **DONE for `xsd:integer` 2026-09-05** (`3914b1a` enumerations, merged; `ed6e41a` INTERVALS + mixed unions via an `IntSet` bucket, squash-merged via PR #106; Konclude 12/12). **Still dropped, deliberately:** interval unions over the five DENSE datatypes (float/double/decimal/date/dateTime — integer discreteness is load-bearing and does NOT transfer, see the `ed6e41a` comment), unions carrying any member no interval set can express (bare datatypes, `DataComplementOf`, other-datatype enumerations — these drop WHOLE and VISIBLY, never narrowed), `DataIntersectionOf`, and qualified data cardinality over a union. Original text: — `DataComplementOf(DataUnionOf(…))`, and `∀` / range / cardinality **over** a union/complement range (dropped).
- [x] **`DataOneOf` enumerations for non-string datatypes** — **DONE 2026-09-04 by #100** (`seed_range_oneof_disjoint`: range and enumeration buckets of the SAME datatype are now compared; integer/decimal/double/float/date/dateTime probes go sat → UNSAT, HermiT-confirmed, string control unmoved). Cross-DATATYPE pairing remains a deliberate sound MISS, pinned by a canary to FLIP rather than delete. Original text: — only `xsd:string` enums are handled. **NOT dropped: converted but the membership clash is never derived, so this one is a SILENT miss, not a reported drop** (measured 2026-09-04, see comment below). integer/decimal/date/dateTime affected.
- [x] **Range-size-aware data cardinality counting** — **DONE** (verified 2026-09-04: `≥3 p` over a 2-value `DataOneOf` yields `unsatisfiable`). Was already fixed by the 2026-06-18 bounded-integer counting work.
  <sub>original text:</sub> ~~Range-size-aware data cardinality counting~~ — e.g. `≥3 p` over a 2-value range ⇒ `⊥` (a concrete-domain counting reasoner; currently only the D4 unsat-clash patterns are caught).

[...]

PR #100, opened 2026-09-04
fix(convert): seed DKey disjointness between range and enumeration buckets (#42 item 2)

micheldumontier:
Closes half of #42 — the `DataOneOf`-for-non-string-datatypes item.

## The defect

`DataHasValue` and facet restrictions lower into a datatype's **range** bucket;
`DataOneOf` lowers into a separate **enumeration** bucket. Disjointness was seeded
*within* each bucket and never *across*, so a value-outside-the-enumeration under a
`∀` never clashed. Same shape as #86 (cross-**datatype** range×range), one level down.

**`xsd:string` was unaffected, and that asymmetry is what found it**: both string forms
lower to one `StrSet` in the `str:` bucket, so `StrSet::disjoint` already compared them.

**Not a dropped construct.** `dropped` was empty on every probe — the axioms converted
fine and the clash was merely never derivable. A *silent* miss, which is why #42 recorded
it as "only `xsd:string` enums are handled".

## Effect

Pinned before/after binaries:

| probe | before | after | Konclude | HermiT |
|---|---|---|---|---|
| integer, value outside the enum | sat | **UNSAT** | unsat | unsat |
| decimal / double / float / dateTime | sat | **UNSAT** | unsat | unsat |
| **date**, value outside the enum | sat | **UNSAT** | *sat (under-report)* | **unsat** |
| every "value INSIDE the enum" control | sat | sat | sat | sat |
| string control (never broken) | UNSAT | UNSAT | unsat | unsat |
| integer range × double enum (cross-datatype) | sat | sat | unsat | unsat |

The last row is a deliberate, pinned **sound MISS** — the seeding is same-datatype only,
which is what keeps `contains` exact (both operands reach their DKey through one
datatype's parse path, so equal values are bit-identical keys). A future fix should
**flip** that canary, never delete it.

On `xsd:date`, HermiT's pair is **discriminating** — outside implies unsat, inside implies
sat — so it is reasoning about the dates rather than refusing them. Konclude's silence is
an eighth recorded under-report.

## Gates

- suite **1892 passed / 0 failed**
- FP=0 net **11 VERIFIED, every closure exact**. The curated corpus is INERT for the DKey
  area by `datatype_value_membership.rs`'s own admission, so this is **non-regression
  only** — the canaries and the HermiT adjudication are the evidence.
- curated fixtures: all 8 answer-identical. `wine`'s byte DIFFER is the documented
  budget-truncation row ordering — the *unchanged* binary produces 3 distinct hashes of
  its own and all 6 runs are content-identical.
- **ORE two-arm sweep, 62 `DataOneOf`-bearing ontologies** (a grep superset — the oneof
  buckets are reachable only from that lowering), arm order alternated:
  **57 IDENTICAL / 3 BOTH_DNF / 2 DIFFER / 0 regressions**, wall **+0.25%**. Both DIFFERs
  adjudicate clean: 3 repeats per arm at a 120 s cap are content-identical, while the
  unchanged binary produces **4 distinct byte hashes of its own** at the sweep's 60 s cap.
- inertness: 24 non-bearing controls, 22 IDENTICAL / 1 BOTH_DNF / 1 DIFFER — and that one
  contains **zero** `DataOneOf`, so the pass provably cannot run on it.

## Sabotage: 4 run, 3 caught

1. delete the seeding — 8 positives fail, controls survive
2. invert the disjointness test — 16 fail (false positives on every negative)
3. `FloatRange::contains` ignoring the inclusivity flags — caught by exactly the
   exclusive/inclusive boundary pair written for it
4. **survived:** dropping the component-reachability gate. That gate is a *cost bound*,
   not an FP guard, so these canaries cannot see it — the same honest limit recorded for
   the #86 split sabotage.

Unflagged: a completeness fix is not opt-in, and `collect_oneof_dkeys` already returns
empty under `RUSTDL_DKEY_ONEOF_SEED=0`, so the flag-off path feeds the new calls nothing.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1504,12 +1504,61 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   `RUSTDL_DKEY_MERGING_GATE=0` — both verified by deliberately breaking them. Spec
   `docs/superpowers/specs/2026-07-30-dkey-nonmerging-component-gate-design.md`.
 
+  **Range × enumeration disjointness (2026-09-04, #42 item 2, unflagged).**
+  `DataHasValue` and facet restrictions lower into a datatype's RANGE bucket;
+  `DataOneOf` lowers into a separate ENUMERATION bucket. Disjointness was seeded
+  WITHIN each bucket and never ACROSS, so `∃p.{5} ⊓ ∀p.{1,2}` did not clash even
+  though 5 ∉ {1,2} — `classify` reported the class satisfiable while HermiT called
+  it unsatisfiable. Same defect shape as #86 (cross-DATATYPE range×range), one
+  level down. **`seed_range_oneof_disjoint` pairs each range bucket with the
+  enumeration bucket of the SAME datatype only**, which is what keeps `contains`
+  exact: both operands reach their DKey through one datatype's parse path, so
+  equal values are bit-identical keys. Cross-datatype pairing would reintroduce
+  the f32/f64 mismatch and is not done; `∃p.{5}^^integer ⊓ ∀p.{1.0,2.0}^^double`
+  therefore stays a sound MISS, pinned by a canary that a future fix should FLIP
+  rather than delete.
+  **`xsd:string` was unaffected, and that asymmetry is what found it** — both
+  string forms lower to one `StrSet` in the `str:` bucket, so `StrSet::disjoint`
+  already compared them. **NOT a dropped construct**: `dropped` was empty on every
+  probe, the axioms converted fine, the clash was merely never derivable — a
+  SILENT miss, which is why #42 recorded it as "only `xsd:string` enums are
+  handled". Effect: integer / decimal / double / float / date / dateTime probes go
+  sat → UNSAT, string control unmoved, every value-INSIDE-the-enumeration control
+  still satisfiable.
+  **Oracles: HermiT confirms every positive; Konclude confirms all but the
+  `xsd:date` one** (it reports `A ⊑ owl:Thing`) — an eighth recorded
+  under-report, not a semantic dispute. HermiT's `xsd:date` pair is
+  DISCRIMINATING (outside ⇒ unsat, inside ⇒ sat), so it is reasoning about the
+  dates rather than refusing them; the neighbouring `seed_cross_bucket_disjoint`
+  note that HermiT throws `UnsupportedDatatypeException` on `xsd:date` is about
+  the CROSS-datatype `date` × `dateTime` construction and does not extend here.
+  **Evidence, with its limit stated:** FP=0 net 11 VERIFIED, closures exact — but
+  the curated corpus is INERT for the DKey area by `datatype_value_membership.rs`'s
+  own admission, so that shows **non-regression only**. The real evidence is 18
+  canaries plus the HermiT adjudication. ORE two-arm sweep over the **62**
+  `DataOneOf`-bearing ontologies (a grep SUPERSET — the oneof buckets are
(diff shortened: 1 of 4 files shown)


micheldumontier:
## Post-merge review of #100 — the FP-safety claim holds, and I confirmed it independently

Reviewed because this lands in the DKey seeding path, it is **FP-direction** (it emits
*more* disjointness), and this is the subsystem `datatype_value_membership.rs` says the
curated corpus is inert for — the one that shipped the `xsd:float`/`xsd:double` false
positive for months behind a green FP=0 net. The PR states that limit itself, which is
the right instinct; this is the independent check.

**Verdict: sound. No changes requested.**

### The seeding direction is right

`seed_range_oneof_disjoint` skips emission when *any* enumeration member lies in the
range (`if set.iter().any(|v| contains(r, v)) { continue }`), so disjointness is asserted
only when **no** member is contained. That is the conservative direction. The safety
property is asymmetric and worth naming: `contains` may over-report membership (fewer
disjointness axioms — a sound MISS) but must never under-report, which would manufacture
a clash. All three implementations handle inclusivity explicitly rather than collapsing to
a `>=`/`<=` pair — `IntegerRange::contains`, `FloatRange::contains` and
`OrdRange::contains` at `data_axioms.rs:507/588/765`.

### I probed the f32-consistency claim rather than take it

"Both operands reach their DKey through one datatype's parse path, so equal values are
bit-identical keys" is the load-bearing sentence, because if the `xsd:float` range bucket
were f32-rounded while its enumeration bucket was not, a boundary value could read as
*outside* the range — exactly the unsafe direction. Confirmed in the source (both round at
f32 then widen: `parse_float_oneof` at `data_axioms.rs:1433`, the float facet parser at
`1970-1981`), and then measured with a probe I did not find in the canary list:

| probe (both `xsd:float`) | rustdl | Konclude |
|---|---|---|
| `∃p.{0.1}` + `∀p.{0.10000000149011612}` — **same f32, different f64** | sat | sat |
| `∃p.{0.1}` + `∀p.{0.25}` — distinct f32 | **unsat** | **unsat** |

The second row is the discriminating control, and it is what makes the first row mean
something: without it, "no unsat" is indistinguishable from "the pass never engaged".
Konclude agrees in both directions, so this is not rustdl grading its own homework.

### Two things I checked because the PR claims them

* The **cross-datatype sound MISS is genuinely pinned**, not just described —
  `a_cross_datatype_pair_is_not_seeded_and_stays_a_sound_miss` with "FLIP this assertion —
  never delete it" in the body. That is the right shape for a deliberate
  under-approximation.
* **The surviving sabotage is correctly characterised.** Dropping the
  component-reachability gate is not caught, and the PR says so and says why: that gate is
  a cost bound, not an FP guard, so canaries asserting *verdicts* structurally cannot see
  it. Same honest limit as #86. Worth carrying forward as a known blind spot — a future
  regression in the gate's *volume* will not fail anything here.

### Gates re-run on current `main`

- DKey canary suite **112 passed / 0 failed** (85 range×oneof + 18 flag defaults + 6
  value-membership + 3 emit-order)
- full suite **1904 passed / 0 failed**

### One note for the release, not a change request

This is unreleased — it landed after the v0.4.26 tag. When it ships, the release note
should carry the PR's own framing rather than a green-tick summary: **the FP=0 net is
non-regression evidence only here**, and the actual evidence is the canaries plus the
HermiT adjudication. On `xsd:date`, HermiT's pair is discriminating while Konclude
under-reports — that is now the eighth recorded Konclude under-report, and it is the
reason the adjudication went to HermiT rather than being called a dispute.

PR #102, opened 2026-09-04
fix(verify): both instrument false-positive mechanisms on the ORE corpus (#87)

### LINKED ISSUE #87: verify-el: all 8 Violated verdicts on ORE are instrument false positives (F4: ⊤⊑C not applied to elements; F5: BoundTripped reported as Violated)
Found by running `verify-el` over all **1,920** ORE ontologies on the post-#81 engine. Of the **366** it can actually check, 358 are `Verified` and **8 are `Violated`** — and **all 8 adjudicate to false positives of the instrument, from two mechanisms**. No new engine defect was found. Full write-up: `docs/benchmarks/2026-08-30-corpus-wide-d10-hunt.md`.

This matters because the crate's headline claim is that a `Violated` is a real engine defect. On the corpus as it stands, **100% of `Violated` verdicts are the instrument's own**, so the two bugs below are what stand between it and being usable unattended.

## F4 — `SubClassOf(owl:Thing, C)` is never applied to model elements

Affects `ore_ont_11522`, `ore_ont_14128`, `ore_ont_14826`, `ore_ont_7270`.

`⊤ ⊑ C` requires every element to be a `C`. `build_model` never propagates it, so the checker reports the very axiom that should have closed the label. The violation count equals the ontology's `owl:Thing`-LHS axiom count **exactly**:

| ontology | `SubClassOf(owl:Thing, …)` axioms | violations |
|---|---:|---:|
| ore_ont_14128 | 2 | 2 |
| ore_ont_7270 | 2 | 2 |
| ore_ont_11522 | 8 | 8 |
| ore_ont_14826 | 19 | 19 |

The reported element is a Tseitin synthetic carrying only itself, e.g. `Element(79)={<synthetic#80>}`.

**The engine is right.** `rustdl subclass-expr <ont> "owl:Thing" "<C>"` returns `yes` on every probed pair (5/5). rustdl simply omits `owl:Thing` from classification output — a reporting convention, not a completeness gap.

Worth flagging beyond this crate: against Konclude these four show `MISSED=8/2/19/2` with `EXTRA=0`, all of the form `Thing ⊑ C`. **Any oracle comparison that counts `Thing` rows will report spurious MISSED** on any ontology asserting `⊤ ⊑ C`. The project has hit this before — 73% of an apparent ~1,795-row gap against Kobayashi-MaRust was the same ⊤-equivalence convention artifact.

## F5 — `BoundTripped` produces `Violated` instead of `Unresolved`

Affects `ore_ont_12317`, `ore_ont_13220`, `ore_ont_15249`, `ore_ont_283` — all 50k–60k classes, all exceeding `Bounds::max_elements = 50_000`:

```
verdict: violated   domain: 50738   violations: 25369
unresolved: ['BoundTripped { bound: "max_elements", limit: Some(50000) }', ...]
```

Violations run to the order of the domain size (25,369 / 20,292 / 54,304 / 92,573). The model was **truncated**, so most axioms cannot be witnessed and the violations are artifacts of the cut.

This is an asymmetry in `owl-dl-cli`'s `fold_build_reasons`: the `Verified` arm downgrades to `Unresolved` whenever build reasons are non-empty (the arm that stops a false all-clear), while the `Violated` arm appends the reasons and **keeps** the verdict. That is right for a complete model with a minor residual, and wrong for `BoundTripped`, where the model is known-incomplete by construction. A truncated model should exit **3** (no verdict), not **2** (real defect).

Note F5 is the cheaper fix and removes half the false positives on its own.

## Coverage, stated as measured

| verdict | count |
|---|---:|
| unresolved (off-fragment/refused) | 1,357 |
| verified | 358 |
| **timeout at 60 s — UNMEASURED, not passing** | **196** |
| violated | 8 |
| I/O or parse error | 1 |

The instrument covers **19%** of the corpus, and the largest single obstacle is the pure-EL fragment restriction rather than defect density.

micheldumontier:
Closes #87.

Of the **366** ORE ontologies `verify-el` can check, 8 reported `Violated` and **all 8
adjudicated to the instrument's own bugs**, not engine gaps — so the crate's headline claim
("a `Violated` is a real engine defect") was, corpus-wide, never once earned. Both mechanisms
are fixed.

## F4 — `SubClassOf(owl:Thing, C)` never reached a Tseitin witness

The saturator emits no fact for a nested existential body and gives its marker an **empty
subsumer set by design**, so the witness element carried only itself and the checker reported
the very `⊤ ⊑ C` axiom that should have closed it. On all four affected ontologies the
violation count equalled the `⊤ ⊑ C` axiom count **exactly**.

**The engine was right throughout**, and the discriminating pair is what shows it: `rustdl
subclass X C` answers `yes`, and the same reproducer with the nesting flattened away verifies
clean. It reproduces in ten lines:

```
SubClassOf(:X ObjectSomeValuesFrom(:r ObjectSomeValuesFrom(:s :A)))   <- nested: Violated
SubClassOf(owl:Thing :C)
```

`FiniteModel::intern` now closes a label containing **no named class** under the ⊤-supers,
collected from both `SubClassOf(⊤, …)` and `EquivalentClasses(⊤, …)` and closed through the
**reported subsumer closure** rather than the axioms — so an engine that failed to derive a
closure step still shows up as a violation instead of being papered over.

## F5 — a truncated model reported `Violated` instead of `Unresolved`

A `BoundTripped` means `build_model` stopped early, so most axioms have no witness left and
the violations are artifacts of the cut: 25369 / 20292 / 54304 / 92573 of them, on the order
of the domain size itself. Now exit **3** (no verdict), not exit 2 (defect found).

## Both fixes are deliberately narrow

Closing **every** label under the ⊤-supers, or downgrading on **any** build reason, each reads
as the natural generalisation — and each would trade this crate's false-**positive** problem
for a false-**negative** one, which is strictly worse for an instrument whose purpose is
finding defects. A named class whose closure lacks a ⊤-super *is* a genuine D10 gap; a
`LabelNotClosed` alongside a violation is still a real disagreement against whatever was
checked.

## A guard test survived its own sabotage

Worth recording, because it nearly shipped. The scoping canary was first written end-to-end,
removing `C` from a named class's element via `test_only_remove_from_label`. That seam runs
**after** interning, so the check fails under either scoping and the test **discriminated
nothing** — it passed the "close every label" sabotage it existed to catch. The rule is now
pinned at unit level on `intern` directly, where it can be observed, and the end-to-end test
was renamed to claim only what it actually checks.

**Sabotage: 5 run, 4 caught first pass, 1 survived and was closed.**

| # | sabotage | result |
|---|---|---|
| S1 | close every label, not just marker-only | **survived**, then caught by the unit pin |
| S2 | drop the subsumer-closure step | caught (1 test) |
| S3 | drop the `EquivalentClasses` arm | caught (1 test) |
| S4 | revert the fix entirely | caught (5 tests) |
| S5 | downgrade on any build reason | caught (negative control) |

## The sentinel is inverted, not deleted

`known_limitations.rs` pins F1/F2/F3 as *current defective behaviour* and its header instructs
that a newly-passing one be updated rather than removed, and be re-checked for whether the
gap is genuinely fixed or merely untriggered by that shape. Answered by measurement: three
further tests cover a 3-deep nesting, the `EquivalentClasses(owl:Thing, …)` spelling, and a
⊤-super reachable only through the closure. **F1 and F3 remain open and still reproduce.**

## Verification

- the 4 F4 ontologies now report `verified`; the 4 F5 ones `unresolved` with exit 3
- suite **1917 passed / 0 failed** (`CARGO_EXIT=0` — captured directly, not through a pipe)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` exit 0
- a two-arm `verify-el` sweep over all 1,920 ORE ontologies is running; I will post the
  verdict-distribution delta as a comment. The risk it covers is a *new* false `Violated`
  from the added labels — the fix only ever adds labels to marker-only elements, so that is
  the direction to check.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1591,6 +1591,35 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   runtime dependency and creates no cycle (nothing in `owl-dl-reasoner`'s manifest names
   `owl-dl-verify`).
 
+> **TWO INSTRUMENT FALSE-POSITIVE MECHANISMS FIXED (2026-09-04, #87), and on the corpus they
+> were 100% of the `Violated` verdicts.** Of the 366 ORE ontologies `verify-el` can check, 8
+> reported `Violated` and **all 8 adjudicated to the instrument's own bugs**, not engine gaps —
+> so the crate's headline claim was, corpus-wide, never once earned.
+> * **F4 (= this doc's F2)** — `⊤ ⊑ C` was never applied to a Tseitin witness. The saturator
+>   gives a marker an EMPTY subsumer set by design, so the witness carried only itself and the
+>   checker reported the very axiom that should have closed it; the violation count equalled the
+>   `⊤ ⊑ C` axiom count EXACTLY on all 4 affected ontologies. `FiniteModel::intern` now closes a
+>   label containing NO named class under the ⊤-supers. **The engine was right throughout**
+>   (`rustdl subclass X C` says `yes`; the same fixture with the nesting flattened verifies).
+> * **F5** — a `BoundTripped` build produced `Violated` instead of `Unresolved`. The model was
+>   TRUNCATED, so violations ran to the order of the domain size (25,369 / 20,292 / 54,304 /
+>   92,573). Now exit 3 (no verdict), not exit 2 (defect found).
+>
+> **BOTH FIXES ARE DELIBERATELY NARROW, and the narrowness is the interesting part.** Closing
+> EVERY label under the ⊤-supers, or downgrading on ANY build reason, would each be defensible
+> one line of reasoning at a time — and each would trade this crate's false-POSITIVE problem for
+> a false-NEGATIVE one, which is strictly worse for an instrument whose purpose is finding
+> defects. A named class whose closure lacks a ⊤-super IS a genuine D10 gap; a `LabelNotClosed`
+> alongside a violation is still a real disagreement.
+>
+> **A GUARD TEST SURVIVED ITS OWN SABOTAGE HERE — the third instance recorded in this file.**
+> The scoping canary was first written end-to-end, removing `C` from a named class's element via
+> `test_only_remove_from_label`. That seam runs AFTER interning, so the check fails under either
+> scoping and the test **discriminated nothing**; it passed the "close every label" sabotage. The
+> rule is now pinned at unit level on `intern` directly, where it can be observed, and the
+> end-to-end test was RENAMED to claim only what it checks. **Sabotage: 5 run, 4 caught first
+> pass, 1 survived and was closed** — run the sabotage before believing a guard.
+
   **Measured coverage, stated as measured, not rounded up.** Inertness (no spurious `Violated` on
   ontologies rustdl is believed complete on) is established on **16 of 20** banner-selected
   pure-EL ORE ontologies; the other **4 are UNMEASURED**, not passing — they hit `timeout`'s
--- a/crates/owl-dl-cli/src/main.rs
+++ b/crates/owl-dl-cli/src/main.rs
@@ -890,6 +890,34 @@ fn fold_build_reasons(
             violations,
             mut unresolved,
         } => {
+            // #87 F5. A `BoundTripped` means `build_model` STOPPED EARLY: the
+            // model is truncated by construction, so most axioms have no
+            // witness left to satisfy them and the violations are artifacts of
+            // the cut rather than evidence of an engine defect. Measured on the
+            // four ORE ontologies that trip `max_elements = 50_000`, the
+            // violation count runs to the order of the domain size itself
+            // (25_369 / 20_292 / 54_304 / 92_573). A truncated model must
+            // therefore yield NO VERDICT (exit 3), not "defect found" (exit 2).
+            //
+            // This is deliberately narrower than "any build reason downgrades a
+            // Violated". Every OTHER reason describes something the builder
+            // finished and could not close (`LabelNotClosed`, `RunDelta`,
+            // `AxiomsDroppedAtConversion`, …), where a violation found
+            // alongside it is still a real disagreement against whatever was
+            // checked — and suppressing those would trade this crate's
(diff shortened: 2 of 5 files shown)


micheldumontier:
Closing as a duplicate — #87 was fixed in parallel and merged as 8bbadb7 / 5a9ea3a / 83bff7a while this was in flight.

I compared the two implementations rather than assuming. **The merged one is a superset of this one on every point but one:**

| | merged | this PR |
|---|---|---|
| F5 narrow to `BoundTripped` only | ✔ `reason_means_model_is_truncated` | ✔ |
| `EquivalentClasses(⊤, …)` arm | ✔ | ✔ |
| ⊤-supers closed through `subsumers_of` | ✔ | ✔ |
| `⊤ ⊑ ∃r.C` — the RULE half | ✔ | �’✗ missing |
| label floor scoping | every label | marker-only labels |

The `⊤ ⊑ ∃r.C` half is a real gap in mine: a label floor cannot build an existential, and I only fixed the label side.

I verified the merged binary independently, on my own harness, on all eight ontologies #87 names — 4 `verified`/exit 0, 4 `unresolved`/exit 3.

Two things from this branch are worth keeping, and I am not re-landing either here:

1. **A corpus-scale two-arm sweep**, which the merged change did not have (its gate was 16 fixtures). Running now against merged `main`; I will post it to #87. It is the gate that can see the specific risk the broader floor introduces — a *new* false `Violated`, or a false all-clear, from labels added at corpus scale.
2. **The scoping difference.** The merged commit states its own residual honestly ("the instrument no longer tests whether the CLOSURE derived `⊤ ⊑ C` for a NAMED class"). Restricting the floor to labels containing no named class recovers that at no cost, and I have a sabotage-verified unit pin for it. Filed separately rather than reopening a merged, deliberate decision.

Worth recording from this branch regardless: my first scoping guard **survived its own sabotage** — it removed a label entry via `test_only_remove_from_label`, which runs *after* interning, so it failed under either scoping and discriminated nothing. The merged commit reports hitting the same class of problem independently ("MY OWN GUARD DID NOT GUARD, FIRST TIME ROUND"). Two of us, same afternoon, same trap.

PR #104, opened 2026-09-05
fix(saturation): a role chain's range must reach the end of the chain (#84)

### LINKED ISSUE #84: Chain-induced range: ObjectPropertyRange does not reach a chain-derived edge between GENERATED witnesses (split from #81)
Split out of #81, whose other reproducer (`cascade.ofn`) is fixed in `4494c4c`. This one is a **different mechanism** and did not move under that fix — the #81 comment claiming it "needs a range folded into a nested witness, exactly as `cascade` does" was wrong.

## Reproducer

`crates/owl-dl-verify/tests/fixtures/chainrange.ofn` (committed, with its oracle):

```
SubObjectPropertyOf(ObjectPropertyChain(:t :u) :r)
ObjectPropertyRange(:r :F)
SubClassOf(:F :G)
SubClassOf(:C ObjectSomeValuesFrom(:t ObjectSomeValuesFrom(:u :A)))
SubClassOf(ObjectSomeValuesFrom(:t ObjectSomeValuesFrom(:u :F)) :D)
```

`C ⊑ D` is entailed:

```
C ⊑ ∃t.∃u.A          -> x∈C has t-succ y, y has u-succ z, z∈A
Chain(t,u) ⊑ r       -> r(x,z)
Range(r,F)           -> z∈F
                     -> y ∈ ∃u.F, so x ∈ ∃t.∃u.F
∃t.∃u.F ⊑ D          -> x∈D
```

| reasoner | result |
|---|---|
| **rustdl** | `[(F,G)]` only — `C ⊑ D` missed, `incomplete: false` |
| HermiT 1.4.3 | `SubClassOf(<C> <D>)` |
| Konclude v0.7.0-1138 | *nothing* — **misses it too** |

## Konclude misses this, so the adjudication needed HermiT

Konclude's silence is normally ambiguous (it under-reports). The **discriminating control** settles it: with the chain removed and `Range` put directly on `u`, Konclude **does** report `C ⊑ D` — so it reports the row when it can derive it.

Probe ladder (`docs/benchmarks/2026-08-29-chain-induced-range-adjudication.md`):

| probe | shape | HermiT | Konclude | rustdl |
|---|---|---|---|---|
| control | `Range(u,F)`, no chain | — | `C ⊑ D` | `C ⊑ D` (post-#81) |
| p1 | chain, no range | — | `C ⊑ D` | `C ⊑ D` |
| p2 | chain + range, `∃r.F ⊑ D` | — | *nothing* | *nothing* |
| p3 | chain + range + `Disjoint(A,F)` | `Nothing ≡ C` | `C ⊑ Thing` | *nothing* |
| p4 | p3 with **asserted** `t(a,b)`, `u(b,c)` | — | inconsistent | inconsistent |
| p5 | p3 with `x : C` (**generated** witnesses) | inconsistent | **consistent** | inconsistent |

p1 shows both apply the chain; p4 shows both propagate a range onto a chain-derived edge between **asserted** individuals. The gap is exactly one combination: **a range propagating onto a chain-derived edge between GENERATED witnesses.**

Note p5: Konclude answers `consistent` where HermiT and rustdl both answer `inconsistent`. That is a stronger claim than the four Konclude under-reports already recorded in this project, and worth reporting upstream.

## Scope in rustdl

rustdl gets **p4 and p5 right** — `abox_saturation` materialises role chains and applies ranges over named individuals. What it misses is **TBox classification**: no chain-induced range reaches a generated witness during saturation.

## Why the obvious fix is unsound

Adding `F` to `effective_ranges[u]` would close this fixture and be **wrong**: a bare `u`-successor with no `t`-predecessor is not an `r`-successor and must not inherit `Range(r)`. This needs a chain-aware derivation rule — when facts `(x,t,y)` and `(y,u,z)` both exist and `t∘u ⊑ r`, propagate `Range(r)` onto `z` — not an extras fold at lowering time.

## Observability

`rustdl verify-el chainrange.ofn` currently exits **3** (`ChainRangeOutOfProfile`) — the checker refuses the shape rather than checking it, so the gap is recorded but not detected by the instrument.

micheldumontier:
Closes #84.

```
Chain(t,u) ⊑ r,  Range(r,F),  C ⊑ ∃t.∃u.A,  ∃t.∃u.F ⊑ D   ⟹   C ⊑ D
```

`classify` reported only `F ⊑ G`, with **`incomplete: false`** — a *silent* miss of the D10
shape. HermiT confirms `C ⊑ D`; **Konclude misses it too**, which is why the adjudication went
to HermiT.

## The issue's prediction was half right

It said this needs "a chain-aware derivation rule, **not** an extras fold at lowering time."
It *is* a fold — but keyed on the **ordered role pair** rather than a single role.

`chain_ranges[(a,b)]` = `effective_ranges[sup]` for every declared `a ∘ b ⊑ sup`, consumed by
the new `nested_extras` at the one site where the composed edge provably exists — lowering
`∃a.∃b.X`:

```
a(x,y) ∧ b(y,z) ∧ a∘b ⊑ sup  ⟹  sup(x,z)  ⟹  z ∈ Range(sup)
```

So `∃a.∃b.X` and `∃a.∃b.(X ⊓ Range(sup))` denote the **same class** — a logical identity, hence
sound and completeness-preserving by construction, exactly as #81's `∃R.B ≡ ∃R.(B ⊓ Range(R))`.
Unflagged for that reason: a differing closure would be a bug, not a trade-off.

**The issue's soundness warning is respected precisely.** Folding into `effective_ranges[u]`
would close the same fixture and be wrong — a bare `u`-successor with no `t`-predecessor is not
a `sup`-successor. The pair key is what avoids that.

## Oracle adjudication — 6 probes, rustdl and HermiT agreeing on all

| probe | rustdl | HermiT |
|---|---|---|
| chain + range, nested | **C ⊑ D** | C ⊑ D |
| range on a **super-role** of the chain head | **C ⊑ D** | C ⊑ D |
| bare `∃u.A`, no `t` step | — | — |
| reversed nesting `∃u.∃t.A` | — | — |
| outer role `w`, not the chain head | — | — |
| no chain axiom | — | — |

## A sabotage survived, and it was the unsound variant

Worth recording. The FP guard used a **top-level** `∃u.A`, which `atomic_existential_rhs`
handles *without ever calling* `nested_extras` — so it guarded the implementation the issue
warns about and was structurally blind to an over-broad fold in the code actually written.

| # | sabotage | result |
|---|---|---|
| S1 | revert the fold | caught (2 tests) |
| S2 | fold into any pair whose second leg is `u` | **survived**, then caught by the new test |
| S3 | key the pair unordered | caught (1 test) |
| S4 | read `role_ranges` instead of `effective_ranges` | caught (1 test) |

Both guards are kept — they cover different implementations of the same mistake.

## The vacuity guard fired, and my first reading of it was wrong

`the_detection_set_has_not_silently_gone_vacuous` failed. That reads as "the fix closed every
gap in the acceptance set." A per-fixture diff against a pinned pre-fix binary showed the truth:
**exactly one of ten fixtures changed** (`chainrange`, the target). The fix is surgical, and
`chainrange` was simply the *last* remaining detection.

Replacement, per the guard's own instruction: **`chaincompose.ofn`** — `t∘u ⊑ r`, `r∘v ⊑ s`,
`Range(s,F)`, a chain composed from two chains. HermiT derives `C ⊑ D`; rustdl reports only
`F ⊑ G`, silently. **This is the documented residual of this very fix**: `chain_ranges` keys on
*declared* pairs, so it sees `(t,u)` and `(r,v)` but never forms the composed `(t,u,v) ⊑ s`.
Closing it needs a transitive closure over chain composition — strictly larger. Its oracle also
records that `build_model` still refuses the shape, so the fixture guards
`classification_matches_oracle` rather than the checker.

## Gates

- suite **1918 passed / 0 failed** (`CARGO_EXIT=0`, captured directly rather than through a pipe)
- FP=0 net **11 VERIFIED, every closure exact** — galen 28007, ore-10080 32356, sio 8904,
  ore-10908 6001, wine 653, pizza 499, ro 158, ore-15672 142, sulo 51, bibtex 16. Its exit 101 is
  the two documented absent fixtures (`alehif-test`, `notgalen`), with **zero closure mismatches**.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` exit 0
- the proof-attribution "mini" pass mirrors the new table, so a proof cannot drift from the
  lowering it explains

A two-arm ORE sweep over the **113** chain+range ontologies plus 22 non-bearing controls is
running; I will post it as a comment. That is the load-bearing gate for a saturator lowering
change, so this should not merge before it reports.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1641,6 +1641,49 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   paths can label one logical nested-existential witness differently and get deduped into two
   separate elements — not shown unsound, but untested against a future concept-level check).
 
+> **CHAIN-INDUCED RANGE FOLDING (2026-09-05, #84, unflagged) — THE SIBLING OF #81, KEYED ON A
+> ROLE PAIR.** `Chain(t,u) ⊑ r` + `Range(r,F)` + `C ⊑ ∃t.∃u.A` + `∃t.∃u.F ⊑ D` entails `C ⊑ D`,
+> and `classify` reported only `F ⊑ G` with **`incomplete: false`** — the D10 shape. `HermiT`
+> confirms; **Konclude misses it too**, which is why the adjudication went to `HermiT`.
+>
+> **The issue predicted this needed "a chain-aware derivation rule, not an extras fold at
+> lowering time". That was half right.** It IS a fold — but keyed on the ORDERED ROLE PAIR
+> rather than a single role. New `chain_ranges[(a,b)]` = `effective_ranges[sup]` for every
+> declared `a ∘ b ⊑ sup`, consumed by `nested_extras` at the one site where the composed edge
+> provably exists, namely lowering `∃a.∃b.X`: `a(x,y) ∧ b(y,z) ∧ a∘b ⊑ sup ⟹ sup(x,z)`. So
+> `∃a.∃b.X` and `∃a.∃b.(X ⊓ Range(sup))` are the SAME class — a logical identity, sound and
+> completeness-preserving by construction, exactly as #81's `∃R.B ≡ ∃R.(B ⊓ Range(R))`.
+> **The issue's soundness warning is respected precisely**: folding into `effective_ranges[u]`
+> would be unsound (a bare `u`-successor with no `t`-predecessor is not a `sup`-successor), and
+> the pair key is what avoids it.
+>
+> **A SABOTAGE SURVIVED, AND IT WAS THE UNSOUND VARIANT.** The FP guard used a **top-level**
+> `∃u.A`, which `atomic_existential_rhs` handles WITHOUT calling `nested_extras` — so it guarded
+> the implementation the issue warns about and was structurally blind to an over-broad fold in
+> the code actually written. Making the lookup match any pair whose SECOND leg is `u` left it
+> green. Closed by `a_nested_pair_whose_outer_role_is_not_the_chain_head_does_not_fire`, which
+> drives that path. **Both are kept — they cover different implementations of the same
+> mistake.** Sabotage: 4 run, 3 caught first pass, 1 survived and was closed.
+>
+> **THE VACUITY GUARD FIRED, CORRECTLY, AND THE FIRST READING OF IT WAS WRONG.**
+> `the_detection_set_has_not_silently_gone_vacuous` failed. That reads as "the fix closed every
+> gap in the acceptance set"; a per-fixture diff against a pinned pre-fix binary showed the
+> truth — **exactly ONE of ten fixtures changed** (`chainrange`, the target), and it was simply
+> the LAST remaining detection. The fix is surgical, not sweeping. **Diff the fixtures before
+> believing a vacuity failure's implied scope.**
+>
+> Replacement detection, per the guard's own instruction: **`chaincompose.ofn`** — `t∘u ⊑ r`,
+> `r∘v ⊑ s`, `Range(s,F)`, a chain COMPOSED FROM TWO CHAINS. `HermiT` derives `C ⊑ D`; rustdl
+> reports only `F ⊑ G`, silently. **This is the documented residual of the #84 fix itself**:
+> `chain_ranges` keys on DECLARED pairs, so it sees `(t,u)` and `(r,v)` but never forms the
+> composed `(t,u,v) ⊑ s`. Closing it needs a transitive closure over chain composition — a
+> strictly larger change. Its oracle also records that `build_model` still REFUSES the shape
+> (`ChainRangeOutOfProfile`), so the fixture guards `classification_matches_oracle`, not the
+> checker.
+>
+> Canaries `crates/owl-dl-reasoner/tests/chain_induced_range.rs` (7). Gates: suite 1918/0;
+> FP=0 net 11 VERIFIED, every closure exact; clippy clean.
+
 > **RANGE FOLDING INTO NESTED EXISTENTIAL WITNESSES (2026-08-29, #81, unflagged) — AND FIVE
 > TESTS THAT WERE PINNING THE BUG.** Two lowering sites in `owl-dl-saturation` dropped a role's
 > `ObjectPropertyRange` when building an existential witness, so `cascade.ofn` returned **ZERO
--- a/crates/owl-dl-reasoner/tests/chain_induced_range.rs
+++ b/crates/owl-dl-reasoner/tests/chain_induced_range.rs
@@ -1,0 +1,191 @@
+//! #84 — a role chain's range must reach the witness at the end of the chain.
+//!
+//! ```text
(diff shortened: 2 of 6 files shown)


micheldumontier:
**Two-arm ORE sweep: 0 regressions, 0 answer changes — and the headline is that nothing moved.**

Pinned binaries, arm order alternated by index, 60 s cap, sequential.

| | frame (113 chain+range) | controls (22 non-bearing) |
|---|---:|---:|
| IDENTICAL | 70 | 20 |
| BOTH_DNF | 32 | 1 |
| DIFFER | 11 | 1 |
| **REGRESSED** | **0** | **0** |

Wall over the both-completed set: **736.6 s → 722.0 s (−1.99%)**.

## All 12 DIFFERs adjudicate clean

Re-run sequentially on an **idle** host at a 300 s cap, compared as closure SETS rather than
bytes (`direct_subsumptions` is a Hasse relation and is *restructured*, not extended, as
reasoning progresses):

- **9 are content-identical** — `gained=0 lost=0`, including the single non-bearing control
  `ore_ont_8911`. Their byte differences were row ordering under truncation.
- **3 move, and all 3 are inside the binary's own run-to-run variance.** Every one reports
  `incomplete: true` in both arms, so the decisive control is the *unchanged* binary against
  itself:

  | ontology | arm difference | **BEFORE vs BEFORE** |
  |---|---|---|
  | `ore_ont_15066` | +12 / −9 | **+14/−20**, **+14/−0** |
  | `ore_ont_205` | +270 / −208 | **+250/−334**, **+310/−319** |
  | `ore_ont_7499` | +339 / −232 | **+221/−275**, **+251/−221** |

  In each case the unchanged binary differs from itself by as much or more than the arms differ
  from each other. Three runs, not two — this repo's own record notes that two cannot establish
  stability on a budget-truncated ontology.

The analyser was **self-tested before use**: `a<b, b<c` closes to 3 pairs, and adding `c<d`
gains exactly 3 and loses 0. It uses per-node BFS, not a memoised `anc(x)` under a
cycle-detection stack — that shape is wrong on cyclic graphs, equivalence groups *are* cycles,
and it previously reported phantom lost entailments on `ore_ont_778`.

## The corpus cannot validate this fix, and that is worth stating plainly

The grep frame selects chain ∧ range co-occurrence. Checking the **precise** precondition — a
chain whose super-role, or a super-role of it, actually carries a range — **61 of the 113
qualify**. So this is not a case of the frame missing the pattern: the precondition holds
broadly across real ORE ontologies, and **not one of them changes any answer**.

What is rare is the *consumer*: a GCI keyed on the range-typed nested existential
(`∃t.∃u.F ⊑ D`). The fold makes the inner witness an `F`; without something that reads that,
no new entailment follows.

So the sweep is **non-regression evidence only**. The evidence for the fix itself is the 7
canaries plus the HermiT adjudication (6 probes, agreeing on all — 2 positives including the
super-role variant, 4 negatives), exactly as with the DKey work where the curated corpus is
inert by its own admission.

Ready to merge from my side.

Issue #99, opened 2026-09-03
Upgrade horned-owl 1.4 -> 3.x and drop the [patch.crates-io] fork (unblocks publishing owl-dl-cli / cargo install rustdl)


micheldumontier:
## Why this exists

`crates/owl-dl-cli` is `publish = false` and `crates/rustdl` ships no `[[bin]]`, and the recorded
reason was that the Manchester-syntax reader was "not yet in upstream `horned-owl`". **That is no
longer true and had not been true for weeks.** phillord/horned-owl#176 merged on 2026-07-11, and
`io::omn` ships in released horned-owl **2.0.0** (2026-07-17) and **3.0.0** (2026-08-10). The stale
claim is corrected in-tree as of e4a857d.

What actually blocks the CLI is narrower and more concrete:

1. this workspace pins `horned-owl = "1.4"` and `io::omn` first appears in **2.0**, so dropping the
   `[patch.crates-io]` fork is a **major-version upgrade, 1.4 -> 3.x**; and
2. one item the fork carries did **not** go upstream.

## What is present upstream vs. fork-only

Probed against docs.rs for 2.1.0 and 3.0.0:

| path rustdl needs | upstream 3.0.0 |
|---|---|
| `io::omn::reader::read` | **present** |
| `io::omn::reader::parse_class_expression` | **present** |
| `io::omn::AsManchester` | **404 - fork-only** |

Upstream merged the reader plus a whole-ontology `io::omn::writer::write`. The fork's per-item
`AsManchester` trait did not land, and `writer::write` cannot substitute: it serialises an entire
ontology, whereas explanations render **one axiom at a time**.

`AsManchester` is reached through `.as_manchester_with_prefixes(&pm)` at **6 call sites in 3 crates**:

- `crates/owl-dl-cli/src/main.rs:1744, 1820, 1919, 1949` (justify / explain output)
- `crates/owl-dl-cli/src/report.rs:85` (HTML report)
- `crates/owl-dl-py/src/explain.rs:14` (Python explain surface)

**Licence constraint:** horned-owl is **LGPL-3.0** and rustdl is Apache-2.0/MIT, so copying the
fork's implementation into this workspace is not clean. The options are to reimplement per-item
Manchester rendering over our own IR, or to upstream `AsManchester` to horned-owl.

## Migration scale (measured 2026-09-03)

- **170 files**, **976** `horned_owl::` references, **49** distinct paths.
[...]

PR #107, opened 2026-09-05
chore: upgrade horned-owl 1.4 -> 3.0 and drop the [patch.crates-io] fork (#99)

micheldumontier:
Refs #99 — steps 1–3. Step 4 turns out to be blocked on something else entirely (below).

## The issue's stated blocker does not exist

#99 records `io::omn::AsManchester` as "404 — fork-only" and scopes the work as reimplementing
Manchester rendering over rustdl's IR, or upstreaming the trait. Reading the 3.0.0 **source**
rather than docs.rs:

```
horned-owl-3.0.0/src/io/omn/mod.rs:5                    pub use writer::{AsManchester, Manchester, write};
horned-owl-3.0.0/src/io/omn/writer/as_manchester.rs:9   pub trait AsManchester<A: ForIRI> {
```

Same design, identical signature, 17 impls — and **upstream is ahead of the fork**, carrying a
PN_LOCAL fix the fork lacks. That scope item is struck.

## The migration is two lines

Zero source changes to any engine crate. Suite **1925 / 0**.

## Three measured costs, none of them in the issue

**1. Parse is 2.7–3.1× slower — and it is parse, not convert.** The fork *also* replaced
horned-owl's `FullIRI` with a fast non-validating rule ("Fast IRI terminals (performance)");
leniency was a side effect. Isolated by subtracting `convert_ms` from wall, min-of-2, idle host:

| ontology | MB | parse 1.4 | parse 3.0 | |
|---|---:|---:|---:|---:|
| ore_ont_7192 | 527 | 3738 ms | 11763 ms | **3.14×** |
| ore_ont_10926 | 558 | 4787 ms | 13158 ms | **2.74×** |
| ore_ont_12898 | 564 | 4971 ms | 13392 ms | **2.69×** |

`convert_ms` is flat throughout; there is no upstream flag to opt out. Impact is
size-proportional and bounded — **+8 s on a ~550 MB file** — and on the 7 largest ontologies
end-to-end at a 60 s cap, **no completion was lost**.

**2. Seven of 1,920 ontologies (0.36%) newly fail to parse**, all with the identical error. **3.0
is right and 1.4 was too permissive**: `ore_ont_10197` carries an IRI containing **U+FFFD** (a
lossy-conversion artifact baked into the corpus file, which is itself valid UTF-8), and U+FFFD
falls outside every `ucschar` range — `A0–D7FF`, `F900–FDCF`, `FDF0–FFEF`, then `10000+`.

**3. `compare-whelk` is removed.** whelk-rs pins `^1.4` (still true at upstream HEAD `a1f7e54`),
and the patch used to unify both sides onto one horned-owl. Cargo won't take two majors as direct
deps of one package and whelk doesn't re-export its horned-owl, so the bench can't parse twice.
`Cargo.toml` records both restoration routes. The lockfile now carries **exactly one** horned-owl
again.

## Publishing stays blocked — for a different reason the issue also misses

Step 4 is not achievable here. Measured:

```
$ cargo publish --dry-run -p owl-dl-cli
error: failed to prepare local package for uploading
Caused by: no matching package named `owl-dl-verify` found
```

`owl-dl-cli` depends on `owl-dl-verify`, itself `publish = false`. Flipping it means first
deciding to publish a diagnostic-only crate — a separate call. The stale horned-owl rationale in
`owl-dl-cli/Cargo.toml` is replaced with this measured one.

## Answer identity

Two-arm parse+convert comparison over **all 1,920** ORE ontologies. `tbox-stats` is parse+convert
with **no reasoning**, so it isolates exactly the layer that changed; arm order alternated,
`convert_ms` stripped as a timing line.

| | count |
|---|---:|
| IDENTICAL converted IR | **1910** |
| **IR content differs** | **0** |
| both fail (pre-existing) | 3 |
| newly fail on 3.0 | 7 |
| newly OK on 3.0 | 0 |

**Zero IR differences** directly answers the issue's own warning that "a parser change can alter
which axioms enter the KB at all". All 8 curated fixtures classify answer-identically (wine's byte
diff is the documented ordering nondeterminism — 3 runs per arm, all content-identical).

## Gates

- suite **1925 passed / 0 failed** (`CARGO_EXIT=0`)
- FP=0 net **11 VERIFIED, every closure exact**; its exit 101 is the two documented absent
  fixtures, with **zero closure mismatches**
- `cargo clippy --workspace --all-targets --all-features` exit 0

One correction worth recording: an earlier note of mine said the upgrade built with "zero source
changes" full stop. That held for the **default** feature set — `--all-features` clippy is what
surfaced the whelk breakage.


micheldumontier:
Parked pending upstream. CI caught an empty-path IRI regression in horned-owl 3.0.0 (`<urn:>`, `<mailto:>`, `<tel:>`) — filed as phillord/horned-owl#292 with the one-line grammar fix. Also: phillord/horned-owl#291 reworks the reader API this branch depends on, so merging now would mean migrating twice. Measurements and rationale recorded on #99; branch stays as-is.

micheldumontier:
Closing as **superseded by #126**, verified by file set rather than by title.

#126 touches **all five** files this PR touches (`Cargo.toml`, `Cargo.lock`, `crates/owl-dl-bench/Cargo.toml`, `crates/owl-dl-bench/src/main.rs`, `crates/owl-dl-cli/Cargo.toml`) and does the same core change — drop `[patch.crates-io]`, `horned-owl` 1.4 → 3.0. It additionally covers six files this one does not, including the `whelk-compare` retirement that the upgrade forces (`whelk-rs` pins `horned-owl = "^1.4"`, so two majors in one graph is an `E0308`) and the Python fixture fix for 3.0 rejecting the bare `<urn:>` prefix.

#126 is `mergeable=true`, `state=clean`, with all CI green (8 success, 1 skipped-by-design).

It also records something this PR's framing does not: the upgrade is **not** fully answer-preserving. 7 of 1,920 ORE ontologies parse under 1.4 and fail under 3.0 — 5 on U+FFFD inside IRIs, 2 on `[`/`]`/`,` in fragments. Both are genuinely invalid per RFC 3987, so 3.0 is correct and 1.4 was lenient, but it is a real capability change and it is documented there rather than discovered later.

Issue #110, opened 2026-09-05
classify silently misses domain- and range-derived subsumptions when the ObjectPropertyDomain/Range filler is conjunctive (D10 shape)


micheldumontier:
## Summary

`ObjectPropertyDomain(r, P ⊓ Q)` + `X ⊑ ∃r.B` entails `X ⊑ P` and `X ⊑ Q`.
`classify` returns **zero rows** with `"incomplete": false` and `"dropped": {}`, under a banner
that certifies the fragment complete. Konclude and HermiT both derive both pairs, and rustdl's
own `subclass` proves them.

This is the D10 shape — a fragment gate certifying COMPLETE while the engine drops the axiom —
so the answer is wrong *and* reported as complete.

## Reproducer

```
Prefix(:=<http://rustdl.test/>)
Prefix(owl:=<http://www.w3.org/2002/07/owl#>)
Ontology(
  Declaration(Class(:P)) Declaration(Class(:Q)) Declaration(Class(:X))
  Declaration(Class(:B)) Declaration(ObjectProperty(:r))
  SubClassOf(:X ObjectSomeValuesFrom(:r :B))
  ObjectPropertyDomain(:r ObjectIntersectionOf(:P :Q))
)
```

| | result |
|---|---|
| `rustdl classify --json` | `direct_subsumptions: []`, `incomplete: false`, `dropped: {}` |
| banner | `# mode: hybrid` / `# fragment: Horn (trust_sat sound by construction; hyper Horn fixpoint is complete)` |
| Konclude v0.7.0-1138 | `X ⊑ P`, `X ⊑ Q` (1,880-byte output — not the 896-byte failure stub) |
| HermiT 1.4.3 | `X ⊑ P`, `X ⊑ Q` |
| `rustdl subclass … X P` | `yes` |

**Discriminating control** — the ATOMIC filler `ObjectPropertyDomain(:r :P)` derives `X ⊑ P`
correctly at the default, and both peers agree. So the cause is the conjunctive filler, not the
domain mechanism.

## Root cause — the tier walk, not the calculus

`subclass` proves the pair, and **no flag on the subsumption path recovers it**:

| flag | rows |
|---|---|
| default | `[]` |
| `RUSTDL_HYPERTABLEAU_TRUST_SAT=0` | `[]` |
| `RUSTDL_HYPERTABLEAU=0` | `[]` |
[...]

PR #111, opened 2026-09-05
docs(verify): NO-GO on widening `verify-el` to Horn, and retract the lever that named it

micheldumontier:
## What this is

A scoping spike with a go/no-go, plus its retraction of the recommendation that motivated it.
**Docs and one new test file; no engine change.**

`docs/benchmarks/2026-09-05-verify-el-cap-is-a-weak-lever.md` closed by naming the fragment as
the binding constraint on `verify-el` coverage — 1,392 of 1,920 (72.5%) `unresolved` against 124
lost to the cap — and recommending the gate be widened from `PureEl` to `Horn`. **That
recommendation is wrong and is retracted here.** A count of refusals is not a count of
recoverable cases.

## The go/no-go

`verify` reaches `Verified` only with **zero** unresolved axioms (`lib.rs:494-520`), so one
refused axiom sinks a whole verdict. A two-list comparison of {constructs that make an ontology
Horn-but-not-EL} against `eval.rs`'s arms finds essentially all of them refused:

- the six concept-level generators (`∃r⁻`, `∀`, `≥n`, `≤n`, nominal, `Self`) — explicit
  `Unresolved` arms
- the twelve axiom-level ones (`Functional`, `InverseFunctional`, `Reflexive`, `Irreflexive`,
  `Asymmetric`, `DisjointObjectProperties`, `DisjointUnion`, the five ABox forms) — `unhandled`
- the inverse-role role-axiom cases — their own guards
- `SymmetricRole` / `InverseObjectProperties` *look* handled but hold only when the role has **no
  edges** — exactly the bare declaration `is_pure_el` already admits

**Exactly two holes**, both the same mechanism (the EL gate demands `Atomic` where `eval.rs`
accepts any concept it can evaluate, and `CE::And` is one): a non-atomic `DisjointClasses`
member, and a conjunctive `ObjectPropertyDomain`/`Range` filler. Grep supersets ≤8 and ≤14, and
**not additive** — a verdict needs *every* departure from `is_el_axiom` to be one of the two.
`ore_ont_10908` is in the 14, is a curated fixture, and classifies at MISSED=0.

**No third hole**, printed rather than eyeballed: `eval_concept` accepts exactly
`Top`/`Bot`/`Atomic`/`And`/`Some(Named)` and `is_el_concept` exactly
`Top`/`Atomic`/`Bot`/`And`/`Some(!inverse)` — identical sets.

## The 1,392 also decomposes, and the second suspected lever is measured out

Full 1,920 scan at a 60 s cap:

| bucket | count |
|---|---:|
| `verified` | 361 |
| refused by the CLI gate | 1,351 — 662 `Horn`, 689 `OutOfFragment` |
| **past the gate, refused by the builder** | **9** |
| timeout — unmeasured | 199 |
| error | 1 |

The builder-refusal half — the population that would need no fragment work at all — is
**9 ontologies** (6 `AxiomsDroppedAtConversion`, 3 `BoundTripped`). `ChainRangeOutOfProfile`,
`LabelNotClosed` and `RunDelta` fire on **zero** ORE ontologies.

## The payoff: a live D10 defect, filed as #110

`ObjectPropertyDomain(r, P ⊓ Q)` + `X ⊑ ∃r.B` entails `X ⊑ P` and `X ⊑ Q`. `classify` returns
**zero rows** with `incomplete: false` and `dropped: {}` under a banner certifying the fragment
complete; **Konclude and HermiT both derive both pairs** and rustdl's own `subclass` proves them.

Root-caused to the tier walk, not the calculus — no subsumption-path flag recovers it
(`TRUST_SAT=0`, `HYPERTABLEAU=0`, `CLASSIFY_VERIFY_REFUTATIONS=1`, `DOMAIN_ABSORPTION=0` all
return zero rows), while `RUSTDL_CLASSIFY_SAME_TIER=1` recovers both. The atomic-filler control
classifies correctly, which isolates the cause.

**Corpus reach measured ZERO** — two-arm run over the 14 candidates, arm order alternated, triple
(rows, unsat, equiv) compared: **13 IDENTICAL, 0 gained, 0 lost, 1 unmeasured**. So
`RUSTDL_CLASSIFY_SAME_TIER`'s corpus-invisible default-OFF justification **stands unqualified**;
this is a second *synthetic* pattern, as `sp11sub` was.

The second commit is a correction worth reading on its own. `ore_ont_12451` showed a **one-sided**
timeout at a 600 s cap (`=0` finished in 230 s, `=1` hit the cap) and was about to be recorded as
a loss. At a 1500 s cap with 2 repeats per arm it completes in **777 s / 775 s against
229 s / 228 s with all four triples identical** — 3.4× slower, same answer. `ore_ont_10080`, the
remaining unmeasured member, DNFs *symmetrically* (600 s / 601 s). **A one-sided timeout at a cap
is a candidate loss, not a loss.** The 3.4× is also the first measurement of that flag's
documented ~2× wall cost on this corpus, which is further reason the #110 fix should not be
"flip the flag".

The defect lives in `classify` and is independent of the widening — `verify-el` cannot even run
on that ontology today. The spike pointed at the shape; the instrument did not detect it.

## Method note kept in the doc

The first draft of this analysis claimed the market was **zero** and carried a `∎`. It was
written over five unread `eval.rs` arms and **two of them were holes**. Enumerate every arm a
universal quantifies over, or do not write the universal.

## Gates

`cargo fmt --all -- --check` clean; `cargo clippy -p owl-dl-verify --all-targets --all-features
-D warnings` clean; `owl-dl-verify` suite **93 passed / 0 failed**. No engine change, so no
FP=0 net or corpus sweep is implicated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1783,6 +1783,19 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   dropping the conjunctive domain filler (which `role_domains` does silently — its own D10 "Bug B"
   comment says so) leaves `X` with NO EL subsumer, hence in `P`'s own tier. **The discriminating
   control is the ATOMIC filler**, `Domain(r, :P)`, which derives `X ⊑ P` correctly at the default.
+  **`ObjectPropertyRange` HAS THE SAME DEFECT AND IS THE LARGER HALF** — `Range(r, P ⊓ Q)` +
+  `X ⊑ ∃r.B` + `∃r.(B ⊓ P) ⊑ W` misses `X ⊑ W` with the identical signature (`subclass` proves it,
+  `TRUST_SAT=0`/`HYPERTABLEAU=0` do not recover it, `SAME_TIER=1` does), and of the 14 ORE
+  candidates **13 are `Range` against 4 `Domain`**. A fix touching only `role_domains` closes half.
+  **`DisjointClasses(A ⊓ B, C)` — the OTHER hole in the same two-list analysis — is CLEAN**
+  (`D` correctly unsatisfiable): disjointness surfaces as an unsat class, which never enters the
+  tier walk, whereas a domain/range fold needs a positive subsumption derived, which does.
+  **AND A NEAR-MISS WORTH THE READ:** the first `Range` probe used `∃r.⊤` and rustdl returned zero
+  rows on the ATOMIC control too — which reads as a pure-EL miss, i.e. far worse. It is not a
+  defect: `∃r.⊤` lowers to a deliberately subsumer-less ⊤-witness so folding `Range(r)` in cannot
+  destroy the `domain(r)` inference the witness exists for — the documented `topwitness.ofn`
+  DESIGN DECISION. Use a concrete filler, not `⊤`, or the control cannot discriminate. Mirror of
+  [[tests-that-pin-the-bug]], applied to choosing a REPRODUCER rather than a fixture.
   **Filed as #110. Corpus reach MEASURED and it is ZERO**: two-arm run (`RUSTDL_CLASSIFY_SAME_TIER`
   0 vs 1, arm order alternated, triple compared) over the 14 conjunctive-`Domain`/`Range`-filler
   ORE ontologies — **13 IDENTICAL, 0 gained, 0 lost, 1 UNMEASURED** (`ore_ont_10080`, which DNFs SYMMETRICALLY at a
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1725,11 +1725,77 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   compute converts 37% of that bucket and **63% is immune** — those are genuinely hard, not
   marginal — buying coverage **18.9% → 21.0%** and finding nothing. Do not spend more compute
   there. **The binding constraint is the FRAGMENT: 1,392 of 1,920 (72.5%) are `unresolved`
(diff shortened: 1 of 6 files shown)


micheldumontier:
**Merge-ordering note — please land #117 first, or resolve in its favour.**

PR #117 (the #110 fix) has landed the engine change this PR's CLAUDE.md text describes as *unfixed*. Two paragraphs here are now falsified by it:

- the `#110` block stating the shape returns `direct_subsumptions: []` with `incomplete: false` and that **only** `RUSTDL_CLASSIFY_SAME_TIER=1` recovers it;
- the `≤14` grep superset — the correct figure is **27**; the scan behind ≤14 had a `break` that fired after the first axiom body matching a broad construct regex, so ontologies whose *first* Domain/Range axiom was non-conjunctive were wrongly excluded.

The "EXACTLY TWO HOLES" market analysis also loses hole 2, which #117 closes. Hole 1 (`DisjointClasses(A ⊓ B, C)`) was separately measured **clean** and needs no fix.

Both branches touch `CLAUDE.md` and `crates/owl-dl-verify/tests/horn_widening_market.rs`. #117's CLAUDE.md entry is **deliberately anchored at this PR's insertion point so the merge conflicts rather than silently interleaving** — two branches quietly producing a self-contradicting design record is exactly the drift this repo's "State of play — 2026-08-18" section documents.

**Resolution, verified rather than assumed:** take #117's version. `a2b64eb` and #117's HEAD differ on that test file only in the third test being retargeted, so nothing is dropped.

Everything else in this PR stands — the `verify-el` widening NO-GO, the two-list comparison, and the 1,920 decomposition are unaffected by #117.

PR #113, opened 2026-09-05
docs: audit #42's datatype residuals — one stale claim, no silent drops

micheldumontier:
Refs #42. Documentation only — no engine change.

All three #42 items are checked, but they were closed by different agents at different times, and
this repo's record has drifted before (#99 alone had two stated blockers that turned out wrong). So
I re-measured the items **and** the residual list they left behind.

## The three items hold

Adjudicated against HermiT:

| item | probe | result |
|---|---|---|
| 1 — nested composites | `∀p.({1} ⊔ {2})` with `∃p.{5}` | `A` unsatisfiable |
| 1 — intervals | `∀p.((≤10) ⊔ (≥100))` with `∃p.{50}` | `A` unsatisfiable |
| 2 — non-string `DataOneOf` | (verified when #100 landed) | — |
| 3 — range-size counting | `≥3 p.({1} ⊔ {2})` | `A` unsatisfiable |

## One residual claim was stale

`DataIntersectionOf` was listed as **still dropped** in ∀/range/cardinality positions. It is not:

```
SubClassOf(:A DataHasValue(:p "50"^^xsd:integer))
SubClassOf(:A DataAllValuesFrom(:p DataIntersectionOf(≥1, ≤10)))
```

→ `A` unsatisfiable, with `dropped` **empty** — the axiom converts and is reasoned about.
**Three-way agreement**: rustdl, HermiT and Konclude all call it unsat. The discriminating negative
control (value `5`, *inside* the intersection) is satisfiable in all three, so the positive is not
vacuous.

## One residual is peer-contested, which makes the visible drop correct

`≥3 p.({1} ⊔ {2})`: **HermiT says unsatisfiable, Konclude says satisfiable** (both measured). Under
this repo's own rule a contested oracle is not an oracle, so rustdl declining to decide it — and
saying so through `dropped` — is defensible rather than a gap to close. This also confirms the
Konclude claim the list already carried.

## The point of the audit: no silent drops

A residual that drops **visibly** is a sound under-approximation a caller can see. One that drops
**silently** (`dropped: {}` *and* `incomplete: false` *and* a wrong answer) is the D10 shape this
repo keeps hitting — and it would be especially easy to miss hiding inside a *deliberate* drop.

All four probed:

| residual | `dropped` | verdict |
|---|---|---|
| dense-datatype interval union (double) | `{unsupported data range: 1}` | visible ✓ |
| union carrying `DataComplementOf` | `{unsupported data range: 1}` | visible ✓ |
| qualified cardinality over a union | `{unsupported data range: 1}` | visible ✓ |
| `DataIntersectionOf` | `{}` | doesn't drop at all ✓ |

**No silent miss was found hiding inside a deliberate one.**

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1678,7 +1678,7 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
     integer interval set can express — bare DATATYPES (`xsd:decimal`), `DataComplementOf`,
     other-datatype enumerations — which drop WHOLE and VISIBLY (`dropped` is non-empty),
     never narrowed; interval unions over the five DENSE datatypes (see the discreteness
-    note above); and `DataIntersectionOf` in these positions.
+    note above).
 
   - **Qualified data CARDINALITY over a union — CLOSED 2026-09-06 (#42).** The list above
     said it "takes a different path", which was true and is now fixed: the three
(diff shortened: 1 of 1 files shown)


micheldumontier:
Rebased onto `main` and resolved the `CLAUDE.md` conflict with #115 (which fixed the very case this branch had documented as a residual).

Two bullets were **superseded rather than merged**, and both would have been wrong to keep:
- "`≥3 p.({1} ⊔ {2})` is peer-CONTESTED, so the visible drop is the right answer" — I had already corrected this once (the spec settles it; HermiT is right), and #115 has now fixed the case outright, so it does not drop at all.
- The still-dropped list's cardinality clause, for the same reason.

Kept, because each is still true and recorded nowhere else: `DataIntersectionOf` is not dropped (three-way agreement plus a discriminating negative control); the silent-drop audit of the four residuals; the OWL 2 spec derivation, now reframed as the *justification* for #115's fix rather than a reason to decline; and the method note that Kobayashi-MaRust cannot break a datatype tie because its `konclude_ht` is a direct port of Konclude's algorithm.

The audit bullet's list of four probed residuals is reworded — two of them turned out not to drop at all.

Issue #114, opened 2026-09-06
classify silently misses domain consequences of the GCI form SubClassOf(ObjectSomeValuesFrom(r, owl:Thing), C) — and certifies it pure-EL


micheldumontier:
## Summary

`SubClassOf(ObjectSomeValuesFrom(:r owl:Thing), C)` is the GCI spelling of `ObjectPropertyDomain(:r, C)`. When `C` is non-atomic, the EL saturator drops it — and unlike #110, **the ontology is certified `pure-EL`**, rustdl's strongest completeness claim.

```
Declaration(Class(:X)) Declaration(Class(:B)) Declaration(Class(:S)) Declaration(Class(:Z))
Declaration(ObjectProperty(:r)) Declaration(ObjectProperty(:s))
SubClassOf(ObjectSomeValuesFrom(:r owl:Thing) ObjectSomeValuesFrom(:s :S))
SubClassOf(:X ObjectSomeValuesFrom(:r :B))
SubClassOf(ObjectSomeValuesFrom(:s :S) :Z)
```

`X ⊑ Z` is entailed: every `X` has an `r`-successor, so `X ⊑ ∃r.⊤ ⊑ ∃s.S ⊑ Z`.

| | verdict |
|---|---|
| **rustdl `classify --json`** | **`direct_subsumptions: []`**, `incomplete: false`, `dropped: {}` |
| banner | **`# mode: pure EL (saturation-only)`** / **`# fragment: pure-EL (trust_sat sound by construction; saturator alone is complete)`** |
| Konclude v0.7.0-1138 | `X ⊑ Z` (1,757 bytes — not the 896-byte stub) |
| Kobayashi-MaRust | `X ⊑ Z`, `dropped: 0` |

## Why this is worse than #110

#110 is certified `Horn`. **This is certified `pure-EL`** — the fragment where the saturator is claimed complete on its own and no tableau runs at all. A silent wrong answer under that banner is the most serious form of the D10 shape.

## Root cause

[...]

PR #117, opened 2026-09-06
fix: decompose a conjunctive ObjectPropertyDomain/Range filler (#110)

### LINKED ISSUE #110: classify silently misses domain- and range-derived subsumptions when the ObjectPropertyDomain/Range filler is conjunctive (D10 shape)

micheldumontier:
Closes #110.

## The defect

`ObjectPropertyDomain(r, P ⊓ Q)` — and the `Range` equivalent — was **silently dropped** by the EL saturator. `collect_el_rules`' arms handled `Bot` and `Atomic` and fell through on `And`, so `classify` returned zero rows with `"incomplete": false` and `"dropped": {}` under a banner certifying the fragment complete. Konclude, HermiT and Kobayashi-MaRust all derive the entailments; rustdl's own `subclass` proves them. That is the D10 shape: **a gate certifies COMPLETE while the engine drops the axiom.**

Prior art, credited: commit `2341ff8` (2026-07-29) identified this as "Bug B" and shipped an interim mitigation — tighten the gates so the axiom is *refused*, routing the ontology to the hybrid path. That moved the miss rather than closing it: its test passes only via `classify_n2`, while the default `classify()` a user gets still returned nothing.

## The fix

`Domain(r, P ⊓ Q) ≡ Domain(r,P) ∧ Domain(r,Q)` is a **logical identity**, so this decomposes rather than approximates. It **reverses** the July mitigation instead of extending it: the ontology becomes genuinely fast-path-safe rather than routed away from a path that could not handle it.

**One shared function, and that is the design.** `pub fn owl_dl_saturation::decompose_role_filler(c, pool, sink) -> bool` pushes each atomic conjunct and reports whether the filler decomposed *completely*. The engine consumes the entries; **both** fragment gates consume the boolean. Gate/engine drift is the mechanism that generates this bug class, and a shared call site makes it unconstructible — the same reasoning that factored out `abox_saturation_inconsistent`.

**Partial decomposition is sound; partial ADMISSION is not.** `Domain(r, P ⊓ ∃s.C)` pushes `P` in the engine (a weaker domain is the safe direction) but is **refused** by both gates — admitting an axiom carrying a conjunct the engine never processed would be a fresh D10 created by the fix for one. A negative canary pins it.

**Why this is safe, and it is not merely "the gate reads the engine's boolean".** `role_domains`/`role_ranges` were **already multi-valued before this change** — `atomic_operands_on_right` pushes multiple heads, and two `Domain` axioms on one role always could — and every consumer iterates the whole `Vec`. Decomposition therefore produces an engine state that was already reachable and already handled.

Also fixed: both **provenance mirrors**, which were `Atomic`-only. Their divergence does not merely empty `axiom_refs` — `mini_effective` drives per-axiom rule-slot delta cursors, so a drift there mis-attributes proofs to the **wrong axiom**.

## Evidence

**Oracles 6/6.** Post-fix rustdl == Konclude == HermiT == KM on named non-`Thing` pair sets across all six probes. Pre-fix returned ∅ with `incomplete: false` on three of them while certifying `Horn`. Konclude judged by content (2414–2537 bytes, never the ~896-byte stub); its silence on the disjunctive probe licensed *affirmatively* via a positive control.

**Sabotage: 6 run, 5 caught, 1 SURVIVED.** The survivor is "the gate re-implements the decomposition locally" — the canaries pin *behaviour*, not the *no-drift property*, which rests on the shared call site alone. Reported, not rounded up. (It is also extensionally undetectable: `ConceptPool::and` flattens nested `And`, drops `Top` and annihilates on `Bot`, so a local `all(matches!(Atomic))` agrees on every interner-constructible input. The risk is future drift, not present incorrectness.)

**Corpus: 45 IDENTICAL / 0 DIFFER / 2 UNMEASURED / 0 regressed / 0 lost entailments** over 25 measured bearing ontologies + 20 controls, triple-compared on the transitive closure. **Zero fragment-routing movers across all 1,920**, measured by gate.

**Corpus reward is measured ZERO.** This is a correctness fix. The FP=0 net's green here is **inertness, not confirmation** — the evidence is the oracle adjudication and the canaries.

> That figure went through three revisions, recorded in `CLAUDE.md` because this project treats design-record drift as a recurring failure: *zero* (right, but over a frame of 14) → *non-zero* (**wrong**: it compared `SAME_TIER=0` vs `=1` within one binary, measuring the flag rather than the fix) → *zero*, measured properly with two pinned binaries. The lesson was already in the file from the label-cache work: **measuring a flag's effect and measuring the ship delta are different questions.**

Frame corrected too: the grep superset is **27**, not the ≤14 previously recorded — the old scan's `break` fired after the first axiom body matching a broad regex.

## Gates

`cargo test --workspace --exclude owl-dl-py` **1960 passed / 0 failed / 85 ignored**; fmt clean; clippy `--workspace --all-targets --all-features -D warnings` clean; doctests clean. `owl-dl-py` excluded: it fails to **link** on undefined CPython symbols on this host — environmental, and this change touches no Python surface.

Commits are **bisect-safe**: the engine fix lands before the gate move, and the intermediate state errs safe (engine derives more, gate still refuses → hybrid).

## Merge note

This branch and `docs/verify-el-horn-widening-no-go` (PR #111) both touch `CLAUDE.md` and `crates/owl-dl-verify/tests/horn_widening_market.rs`. The CLAUDE.md entry here is **deliberately anchored at the sibling's insertion point so the merge conflicts rather than silently interleaving** — PR #111 carries a `#110`-is-unfixed paragraph and a `≤14` frame that this branch falsifies.

**Resolve in favour of this branch**, verified rather than assumed: `a2b64eb` and this HEAD differ on that test file only in the third test being retargeted, so nothing is dropped.

## New infrastructure

`scripts/closure-diff.py` — the transitive-closure comparator behind the "0 lost entailments" result, committed so the claim is reproducible. Fresh per-node BFS, equivalence groups expanded as cliques, no shared memoisation, with a `--self-test` that demonstrates a buggy memoised DFS **disagreeing** with it (`closure(Q)`: buggy `{P}` vs correct `{P,X,Y}`). This repo has already produced a false "3 lost entailments" finding from a memoised comparator.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1730,6 +1730,73 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
   *increased* unresolved by 31 — several slow ontologies were off-fragment all along, so part of
   the timeout bucket was never checkable. See
   `docs/benchmarks/2026-09-05-verify-el-cap-is-a-weak-lever.md`.
+  **THE SPIKE'S PAYOFF WAS A LIVE D10 DEFECT IN THE SATURATOR, AND IT IS NOW FIXED (#110,
+  2026-09-06, unflagged).** `ObjectPropertyDomain(r, P ⊓ Q)` + `X ⊑ ∃r.B` entails `X ⊑ P` and
+  `X ⊑ Q`. `classify --json` returned **`direct_subsumptions: []` with `incomplete: false` and
+  `dropped: {}`** while certifying `# fragment: Horn` — the gate promising completeness over an
+  axiom the EL saturator silently dropped, the D10 shape, and worse than a DNF. `ObjectPropertyRange`
+  is the same defect and the larger half. **Fixed by DECOMPOSITION, which is a logical identity**
+  (`Domain(r, P ⊓ Q) ≡ Domain(r,P) ∧ Domain(r,Q)`), hence sound and completeness-preserving by
+  construction. **Wherever this defect is recorded as UNFIXED with `RUSTDL_CLASSIFY_SAME_TIER=1` as
+  the only recovery, that record is superseded here**: both pairs now come out at the DEFAULT, and
+  it was never only a tier-walk story.
+  **The design point is the SHARED CALL SITE, not the rule.** One
+  `pub fn owl_dl_saturation::decompose_role_filler(c, pool, sink) -> bool` pushes every atomic
+  conjunct *and* returns whether the filler decomposed COMPLETELY; the engine consumes the entries
+  and **both** fragment gates consume the boolean (`classify.rs::is_processed_role_filler`, née
+  `is_atomic_or_trivial_concept`, at all four call sites — `Bot` re-admitted explicitly there
+  because `poisoned_roles` handles it before the decomposer is reached). Gate/engine drift is the
+  mechanism that GENERATES this bug class, so the fix makes it structurally impossible rather than
+  re-aligning two predicates that happen to agree today.
+  **Partial decomposition is sound; partial ADMISSION is not.** `Domain(r, P ⊓ ∃s.S)` still
+  contributes its `P` conjunct — a weaker domain is the safe direction — but the ontology must NOT
+  be certified EL-complete on that basis, since the `∃s.S` half is genuinely dropped. A negative
+  canary pins exactly that asymmetry.
+  **EVIDENCE, and the corpus is not part of it.** Oracles: post-fix rustdl == Konclude == HermiT ==
+  KM on **6 of 6** probes (pre-registered comparison of named non-`Thing` pair sets); the pinned
+  pre-fix binary returns **∅ with `incomplete: false` on three of them while certifying `Horn`**.
+  Sabotage: **6 run, 5 caught, 1 SURVIVED.** The survivor is the mutation that makes the gate
+  re-implement the decomposition LOCALLY instead of calling the shared function — invisible to the
+  canaries because they pin BEHAVIOUR, not the no-drift property. It is bounded, not unknown
+  (`ConceptPool::and` flattens nested `And` at intern time and there is exactly ONE
+  `ConceptExpr::And` construction site workspace-wide, so the two predicates are equal on every
+  interner-constructible input), so the residual is an untested CROSS-CRATE COUPLING and the
+  no-drift property rests on the shared call site alone. **Report the survivor; do not round it up
+  to 5 of 5.**
+  **CORPUS REWARD IS MEASURED ZERO.** Frame = the **27** ORE ontologies authoring a conjunctive
+  `Domain`/`Range` filler. Two pinned arms: **45 IDENTICAL / 0 DIFFER / 2 UNMEASURED / 0 `ok → dnf`
+  / 0 lost entailments** over 25 measured bearing + 20 controls, and **0 fragment-routing movers
+  across all 1,920** by GATE probe. So the FP=0 net's green here is **INERTNESS, not confirmation**
+  — the evidence is the oracle adjudication and the canaries.
+  **The fix has TWO SEPARABLE EFFECTS and the gate sweep bounds only one.** A ROUTING effect (the
+  gate admits the axiom, moving an ontology to the fast path only if that axiom was its sole
+  blocker) and a DERIVATION effect (the saturator decomposes regardless of the gate).
+  `Domain(r, P ⊓ ∃s.S)` separates them — no routing move, 0 → 2 rows. Stopping at "0 movers" would
+  have left the larger half unmeasured.
+  **THE SUPERSET IS 27, NOT ≤14.** Wherever ≤14 is recorded for this shape it is stale: that scan's
+  `break` fired after the first axiom body matching a broad construct regex, so any ontology whose
+  FIRST `Domain`/`Range` axiom was non-conjunctive was excluded. Re-derived by two agreeing
+  instruments.
+  **The two-hole `verify-el` widening market loses hole 2.** `Domain(r, P ⊓ Q)` is no longer a
+  gate-vs-checker disagreement — it decomposes to EL and reports `PureEl`, a stronger outcome than
+  the hole count implied. Hole 1 (`DisjointClasses(A ⊓ B, C)`) is untouched by #110 and was
+  measured clean separately, outside this branch's evidence — attribute it there, not here.
+  `crates/owl-dl-verify/tests/horn_widening_market.rs`'s second test is retargeted to an
+  EXISTENTIAL domain filler, which stays out of EL for a DESIGN DECISION rather than a defect
+  (`decompose_role_filler` pushes `ClassId`s and `∃s.S` has none, so admitting it needs a
+  structurally different rule, not a wider decomposer). `ObjectUnionOf` was tried first and
+  measures `OutOfFragment`, not `Horn`. See [[tests-that-pin-the-bug]].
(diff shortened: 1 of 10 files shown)


micheldumontier:
Closing as **superseded**: `main` moved during this branch's execution and #110 was fixed there by #112 (`ed2ab5f`), with #114 fixed by #116 (`c708286`). This branch's engine change is a parallel implementation of the same identity-based decomposition, so re-litigating it would be churn.

**My process error, recorded because it is the interesting part.** I re-verified the premise against `main` *before* planning — the rule this repo has a standing section about — and then did not re-check across a multi-hour execution. Branching from `0abc21b` and reviewing five tasks against it was correct at the time and stale by the end. A `git fetch` before each task's dispatch would have caught it at Task 2.

**The branch was not wasted: it surfaced two residuals in #112's fix, both verified against peers and now filed.**

- **#118 — `prove` cannot attribute the derivation.** #112 changed the real Pass 1 but not the two provenance mirrors, which still accept `Atomic` only. With the discriminating control: the atomic filler gives `(Domain(sub)) ⊢ X SubClassOf P  [axiom[1]]`, the conjunctive one gives the same step with **no axiom reference**. Correct answer, missing justification. The range mirror is the worse half — `mini_effective` drives per-axiom rule-slot delta cursors, so drift mis-attributes proofs to the *wrong* axiom rather than merely emptying `axiom_refs`.
- **#119 — a partly-atomic filler is still dropped whole.** `all_atomic_conjuncts` is all-or-nothing (`?` returns `None` on the first non-atomic conjunct), so `Domain(r, P ⊓ ∃s.S)` yields nothing on `main` while **Konclude and Kobayashi-MaRust both derive `X ⊑ P`**.

A focused follow-up rebased on `main` will carry only those two fixes, their canaries, and `scripts/closure-diff.py` — the per-node-BFS transitive-closure comparator, committed so closure claims are reproducible (this repo has already produced a false "3 lost entailments" finding from a memoised comparator).

Deliberately **not** carried over: the fragment-gate move. It would restore pure-EL fast-path eligibility for these ontologies, but corpus reach is measured zero, so it is a performance change with no evidence behind it. Instead the follow-up pins a canary that the gate **refuses** a partly-decomposable filler, so that widening it later cannot silently admit an axiom the engine only partly processed.

Also worth keeping from this branch's evidence, independent of the implementation: three peer reasoners agree 6/6 on the fixed shapes; and the corpus reward for #110-class fixes is **measured zero** over the corrected 27-ontology frame (not the ≤14 previously recorded), with 0 lost entailments and 0 fragment-routing movers across all 1,920.

Branch left in place at `fix/conjunctive-domain-range-filler-110` for reference; nothing here needs merging.

PR #132, opened 2026-09-08
docs: VarCap is measured out — 11 Horn-bodied refusals, all inert (WS2 negative)

micheldumontier:
Two commits that were sitting **unpushed on a local `main`** and would have been lost with the machine. `origin/main` still carried the superseded conclusion, so this is a correction, not just an addition.

## What changed in the conclusion

`main` says `VarCap` "**is the live residual** here at 39 ontologies" — a clause-discarding refusal (`MAX_BODY_VARS`) left open as the last addressable item in that census. Measured, it is **inert**, and the route there corrects two things that had been inferred rather than measured:

| | before | measured |
|---|---|---|
| refusing ontologies | 39 | **37** |
| expected to be all disjunctive | assumed | **11 are ALL-HORN** (`disjunctive=0`) — refutes the expectation |
| variables needed | spread 9/11/12/16/25/133 | all 11 refuse at **exactly 9** — the cap is off by ONE for this set |
| answer changes at cap 8 vs 16 | unknown | **0 of 4 measurable; 7 DNF symmetrically in both arms** |

So the shape is real and the reward is zero: with no answer to change, there is nothing for a withheld clause to alter. The claim is deliberately bounded as **0 of 4**, not 0 of 11, because the other 7 produce no verdict in either arm.

Also recorded: `ore_ont_7775` is both one of the 11 **and** one of the three completers that cap 16 is already documented to destroy — so the naive fix trades known recoveries for known regressions.

## The instrument note is the transferable part

The first positive control **did not fire**: the fixture was pure-EL, so `classify` took the saturation fast path and never reached the hyper engine. A census run at that point would have read "0 addressable" from a silent instrument and looked like a clean negative. Adding one `∀` fixed it.

## Why this is a separate PR

The 90-line benchmark doc exists nowhere but this branch. Kept off #130 (which is three unrelated `CLAUDE.md` fact corrections) so each can be reviewed and reverted on its own.

Documentation only — no code, no behaviour change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1118,10 +1118,23 @@ Data flows: `horned-owl` parse → `owl-dl-core` (IR + preprocessing) →
     NEGATIVE one, which includes an ordinary `EquivalentClasses(C, ObjectHasSelf(r))`: the
     `⊒` direction puts `Self` in the antecedent, and NNF makes it negative. So a grep for
     `ObjectComplementOf(ObjectHasSelf(…))` gives the right answer for the wrong reason;
-    re-scanned for the corrected pattern, ORE still has **0 of 9**. **`VarCap` is the live
-    residual here at 39 ontologies** — a clause silently discarded for exceeding
-    `MAX_BODY_VARS`, the same shape `NotTree` had — recorded, not queued, since raising that
-    cap is a measured hard stop. Two instruments disagreeing on COUNTS (`ro` 2 vs 6) is
+    re-scanned for the corrected pattern, ORE still has **0 of 9**. **`VarCap` WAS the live
+    residual here at 39 ontologies — MEASURED OUT 2026-09-06.** Re-censused at **37** refusing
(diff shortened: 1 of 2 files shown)


micheldumontier:
Closing as **redundant** — this was opened on a mistaken premise.

I found commits `7322b28` / `ded32b6` sitting on a local `main` that was two commits behind and three ahead of `origin/main`, and read that as work existing only on one machine. It is not: **#126 already carries both of them** as its first two commits, pushed on `fix/horned-owl-3x-drop-fork`. Nothing was ever at risk.

The content here is byte-identical to what #126 carries (the benchmark doc hashes the same on both branches), so merging this would either conflict with #126 or apply as a no-op depending on order. #126 is green and ready; it should carry them.

What was genuinely true and remains so: `origin/main` still has the **superseded** VarCap text calling it "the live residual here at 39 ontologies", and the 90-line benchmark doc is absent from `main`. Both are fixed by #126, not by this.

No content is lost by closing this — the branch stays pushed if it is ever wanted.

Issue #128, opened 2026-09-07
Missing expected equivalent classes (symmetric property with a property chain)


PiotrNowara:
Hi,

This project looks really useful!

I tried `rustdl` with a small ontology featuring a combination of a property chain and symmetric property. Unfortunately I don't get expected results: classes `DiscomfortT1a` and `DiscomfortT1b` are not inferred as equivalent (this is inferred correctly by HermiT). I'm attaching the OWL file.

```
Class: DiscomfortT1a

    EquivalentTo: 
        Discomfort
         and (coOccursWith some Yawning)
         and (precedes some Pain)
    
    
Class: DiscomfortT1b

    EquivalentTo: 
        Discomfort
         and (coOccursWith some 
            (Yawning
             and (precedes some Pain)))
```

Below a screenshot with HermiT explanation:
<img width="920" height="193" alt="Image" src="https://github.com/user-attachments/assets/db61af02-e30c-4339-a97a-21d12724c05c" />

Best,
Piotr

[epi_corrected_debug_minimal.ow.l.txt](https://github.com/user-attachments/files/31925105/epi_corrected_debug_minimal.ow.l.txt)

micheldumontier:
Thanks for the report — this is a real bug, it's confirmed, and your reproducer localised it precisely. Both peer reasoners agree with you: HermiT emits `EquivalentClasses(:DiscomfortT1a :DiscomfortT1b)` and so does Konclude v0.7.0-1138.

## What's happening

`DiscomfortT1b ⊑ DiscomfortT1a` **is** derived — that direction needs only the property chain. The missing direction is `DiscomfortT1a ⊑ DiscomfortT1b`, which needs **symmetry composed with the chain**: from `coOccursWith(x,y)` and `precedes(x,z)`, symmetry gives `coOccursWith(y,x)`, and then `coOccursWith ∘ precedes ⊑ precedes` gives `precedes(y,z)` — which is what puts the `Pain`-successor on the `Yawning` witness.

Minimised to five axioms, with your annotations, transitivity and domain/range all stripped:

```
SymmetricObjectProperty(p)
SubObjectPropertyOf(ObjectPropertyChain(p q) q)
A ≡ ∃p.Y ⊓ ∃q.Z
B ≡ ∃p.(Y ⊓ ∃q.Z)
```

| shape | `A ⊑ B` | `B ⊑ A` |
|---|---|---|
| symmetric + chain | **no** | yes |
| chain only | no | yes |
[...]

Issue #131, opened 2026-09-08
ObjectPropertyRange on a chain's second leg or super role blocks the chain-derived subsumption


micheldumontier:
Split out of #128, which turned out to expose **two** independent gaps. This is the second one; the symmetry-composed-with-chain gap stays on #128.

## Summary

Adding `ObjectPropertyRange` to a role-chain's **second leg** or to its **super role** stops a chain-derived subsumption from being found. No symmetry is involved — this reproduces with an explicitly named inverse, so it is not #128's mechanism.

## Reproducer

```
Prefix(:=<http://ex.org/>)
Ontology(<http://ex.org/o>
Declaration(Class(:A)) Declaration(Class(:B)) Declaration(Class(:Y)) Declaration(Class(:Z)) Declaration(Class(:Proc))
Declaration(ObjectProperty(:p)) Declaration(ObjectProperty(:q)) Declaration(ObjectProperty(:pi))
InverseObjectProperties(:p :pi)
SubObjectPropertyOf(ObjectPropertyChain(:pi :q) :q)
SubObjectPropertyOf(ObjectPropertyChain(:p :q) :q)
ObjectPropertyRange(:q :Proc)
EquivalentClasses(:A ObjectIntersectionOf(:D ObjectSomeValuesFrom(:p :Y) ObjectSomeValuesFrom(:q :Z)))
EquivalentClasses(:B ObjectIntersectionOf(:D ObjectSomeValuesFrom(:p ObjectIntersectionOf(:Y ObjectSomeValuesFrom(:q :Z)))))
)
```

`A ≡ B` is entailed and **HermiT derives `EquivalentClasses(A B)`**. rustdl answers `A ⊑ B` = `no`. Removing the single `ObjectPropertyRange(:q :Proc)` line flips rustdl to `yes`.

## Localisation

With a chain `pi ∘ q ⊑ s` and a distinct super role `s`:

| range placed on | `A ⊑ B` |
|---|---|
| nothing | **yes** |
| second leg (`q`) | **no** |
| super role (`s`) | **no** |
| first leg (`p`) | yes |

Also isolated on the simpler self-super form (`pi ∘ q ⊑ q`): **`Range` breaks it, `Domain` does not**, and unrelated `SubClassOf(… Process)` axioms do not.

## Suspected mechanism (hypothesis, not verified)

[...]

Issue #124, opened 2026-09-07
completeness_guaranteed()'s Horn arm is a false certificate: classify misses an oracle-confirmed subsumption while disjoint --json reports incomplete: false


micheldumontier:
## Summary

`Classification::completeness_guaranteed` (`crates/owl-dl-reasoner/src/classify.rs:912`) returns true for `PureEl | Horn` when `timed_out_pairs == 0`. On a `Horn`-classified ontology `classify` can miss a subsumption **both peer oracles derive**, and the certificate still reads clean — so the miss is surfaced nowhere.

This is the D10 shape (wrong answer AND reported complete) **one level up from the fragment gate, at the public API**. `CLAUDE.md` already records the underlying justification as false-as-implemented for the D10 Horn-shortcircuit — *"The 'hyper Horn fixpoint is complete on Horn' justification was false-as-implemented: the shortcircuit ran the EL saturator, not the hyper fixpoint"* — but `completeness_guaranteed()` still spends that same claim.

## Reproducer

```
Prefix(:=<http://ex.org/>)
Ontology(<http://ex.org/t>
Declaration(Class(:X)) Declaration(Class(:B)) Declaration(Class(:Z)) Declaration(Class(:W))
Declaration(ObjectProperty(:r)) Declaration(ObjectProperty(:s))
SubClassOf(:X ObjectSomeValuesFrom(:r :B))
ObjectPropertyDomain(:r ObjectSomeValuesFrom(:s :Z))
SubClassOf(ObjectSomeValuesFrom(:s :Z) :W)
)
```

`X ⊑ W` is entailed. Measured on a freshly built `main` (`a04c548`, binary `5e8c0bad…`):

| surface | result |
|---|---|
| `classify --json` | `direct_subsumptions: []`, **`incomplete: false`**, `dropped: {}` |
| `subclass X W` | **`yes`** |
| `disjoint --json` | **`incomplete: false`** |
| `# fragment:` banner | `Horn (trust_sat sound by construction; hyper Horn fixpoint is complete)` |

**Both oracles confirm `X ⊑ W`:**
- Konclude v0.7.0-1138 — reports `X <= W` (4 `SubClassOf` blocks, judged from content not exit code; 1712-byte output, well above the ~896-byte stub)
- HermiT 1.4.3 via robot 1.9.10 `reason --reasoner hermit --axiom-generators SubClass` — `SubClassOf(<http://ex.org/X> <http://ex.org/W>)`

[...]

PR #133, opened 2026-09-08
fix(tableau): a chain leg must be walkable through an inverse-equivalent edge (#128)

micheldumontier:
Refs #128 — deliberately NOT `Fixes`. This repairs the calculus, but default `classify` still misses the reporter's entailment (see *What this does NOT fix*), so #128 must stay open until the wedge side lands.

## The bug

`apply_role_chains` resolved each chain position against raw edge labels **and directions** — a `Named(r)` leg read only `r`-labelled out-edges — so a chain whose leg had to be traversed *backwards* never fired, even when an `InverseObjectProperties` declaration made that hop an identity rather than an assumption. Since `SymmetricObjectProperty(r)` is lowered to `InverseObjectProperties(r, r)`, symmetric roles were the common way in:

```
Symmetric(coOccursWith),  coOccursWith ∘ precedes ⊑ precedes
DiscomfortT1a ≡ Discomfort ⊓ ∃coOccursWith.Yawning ⊓ ∃precedes.Pain
DiscomfortT1b ≡ Discomfort ⊓ ∃coOccursWith.(Yawning ⊓ ∃precedes.Pain)
```

HermiT reports these equivalent. rustdl reported only `T1b ⊑ T1a` — the direction that reads the chain forwards and never needed symmetry. `T1a ⊑ T1b` requires the chain to fire **at the `Yawning` witness** `y`, through `coOccursWith(y,x)`, which exists purely because `coOccursWith` is symmetric.

`edge_satisfies`, the concept-level matcher, had been accepting that cross-polarity match all along for `∃`/`∀`. The chain rule was the one component that disagreed about which edges exist.

## Diagnosis, with a control

Replacing `Symmetric` with an explicit `inverse(coOccursWith) o precedes` chain answered correctly *before* this change — so the `Inverse` chain position always worked, and it was specifically the symmetry-derived hop that was missing. Sub-role legs also already worked (sub-role edges are materialised up the hierarchy before chains run), which isolates inverse-equivalence as the sole gap.

The defect is not symmetry-specific: the general two-role `InverseObjectProperties(r, s)` case failed identically, and is covered here.

## The change

Chain legs now also accept an opposite-direction edge whose role is a declared inverse of the wanted one, using the same `are_declared_inverses` predicate `edge_satisfies` uses — so the two matchers now agree. `r ≡ s⁻` is an equivalence, so the extra hop derives nothing a model could refuse. A role with no inverse declaration is untouched, and the predicate short-circuits on an empty pair set, so an ontology with no inverse declarations pays nothing.

Net effect on `rules.rs` is a small extraction: the four hand-inlined edge scans collapse into one `chain_leg_targets` helper used by both legs.

## Verification

The change only ever *adds* derived edges, so false positives were the entire risk direction.

- **Corpus soundness net** (`run-soundness-diff.sh`) — run on this branch and on a pristine-`HEAD` worktree sharing the same oracles. `[fp0]` manifests, per-subsumption MISSED/FP lists and failure sets all diff clean. `sio` 8904=8904, `wine` 653=653, `ro` 158=158, `sulo` 51=51, all FP=0/MISSED=0; `family` inconsistency detected. Those four exactly reproduce the reference values in `CLAUDE.md` from oracles generated here by a different Konclude build — the pipeline-validation control.
- **ORE 2015 paired A/B** — scoped to the 628 pool ontologies where the fix can fire (a chain leg, explicit or from transitivity, meeting an inverse-equivalence on the **same role**). Re-run in full on the final code, both binaries rebuilt on the horned-owl 3.0 base. 610 swept, **604 SAME / 6 DIFF**, 463 with both binaries completing:

  | metric | baseline | this branch |
  | --- | --- | --- |
  | rustdl subsumptions | 6,423,849 | 6,423,849 |
  | oracle subsumptions | 6,428,115 | 6,428,115 |
  | false positives | **0** | **0** |
  | MISSED | 4,266 | 4,265 |

  All 6 DIFFs are measurement artifacts, individually adjudicated: 4 TIMEOUT-vs-completed flips at the 90 s cap in *both* directions (3 favouring this branch, 1 the baseline); `ore_ont_12451` exhausting the harness's own 32 GB `ulimit -v` on **both** binaries (not a code panic — only which limit landed first differed); and `ore_ont_7532`'s lone `MISSED 2 → 1`, which **is not a recovery** — at a 300 s cap both binaries report `10552/10553 MISSED=1` identically, 3/3 runs, so the 90 s row was the baseline losing a pair to the 200 ms budget under contention. That last one is the MISSED column's 4,266-vs-4,265: noise, not a gain. Net: no attributable behavioural difference on the ORE trigger set.
- **Unit** — 8 tests in `chain_symmetric_leg.rs`, all out of the EL fragment, including three FP guards and Fable's sub-role counterexample. **Mutation-checked**: both negatives fail under `true || are_declared_inverses(..)` and under a second mutation that walks `super_roles` without checking the inverse partner. See *Review corrections* below — the first revision's guards were vacuous.
- fmt clean, clippy `-D warnings` clean, workspace tests and doctests green.

## Review corrections (please read before approving)

An adversarial review found two defects in the **first** revision of this PR. Both are fixed; recording them because one of them invalidated a verification claim I had made in the commit message, this body, the CHANGELOG and on #128.

**1. The two "FP guards" were vacuous.** Both fixtures declared no symmetry and no inverse pair, making them pure EL — `is_subclass_of` returns the saturation-closure verdict and `chain_leg_targets` never executes. Mutating the reverse read to `true || are_declared_inverses(..)`, the exact unsound fix they were claimed to catch, left **all six tests passing**.

Two generalisable lessons, now written into the test file:

- **Every tableau fixture must be out of the EL fragment.** An EL-derivable subsumption is answered by the saturator, which handles role hierarchy and chains fine — so an EL fixture pins the saturator, not the code under test. This cuts both ways: an EL-derivable *positive* is masked identically, so a "chain + sub-role works" control proves nothing about the chain rule. (That mistake also weakened my localisation on #138, corrected there.)
- **Leaving EL is harder than it looks.** `BareRoleDecls` admits an inverse/symmetry declaration over a role no axiom can *read* as semantically inert, so a pair over unused roles keeps the ontology pure EL. It must be over an **observable** role. My first attempt at fixing the guard failed for exactly this reason.

Acceptance criterion now stated in the file: **a test that survives that mutation is not a guard.**

**2. The doc comment asserted a mechanism that does not exist**, and it hid a real miss. I wrote that `apply_role_rules` materialises super-role edges so chains need not consult the hierarchy. It does not — it adds concept *labels*, and `add_edge` records only the exact role; the tableau consults the hierarchy lazily via `edge_satisfies`, which the chain rule did not use. Consequence, one step from the #128 fixture: `q ⊑ co` + `Symmetric(co)` + `co ∘ p ⊑ p` missed its entailment, and no EL closure can rescue it because symmetry is out of fragment.

Fixed by composing both reads with the hierarchy (see the updated `chain_leg_targets`), pinned by `a_subrole_of_a_symmetric_role_fires_the_chain` with `a_subrole_of_an_unrelated_role_does_not_fire_the_chain` as its FP guard.

The review explicitly cleared, having tried to break them: soundness of the reverse read across both polarities and self-loops; dep-set correctness (`enumerate()` precedes `filter()`, and `add_edge_inner` writes identical deps to both lists); `chain_edge_already_present` being a missed optimisation only; termination.

## What this does NOT fix

**Default `classify` still misses the reporter's entailment; `is_subclass_of` gets it right.** `classify` never reaches this rule on that shape — the hypertableau wedge decides the pair first, and the wedge can traverse a symmetric edge backwards only when the role hierarchy is threaded in, which happens solely under `RUSTDL_CLASSIFY_SAME_TIER=1` ("DEFAULT OFF — sound (FP=0) but corpus-invisible and ~2× wall"). **#128 is a user-reported counterexample to "corpus-invisible", so that default is now a live question.** Pinned visibly by `default_classify_still_misses_the_symmetric_chain_leg` rather than `#[ignore]`d, per the convention in `owl-dl-verify/tests/known_limitations.rs`.

Two further gaps found and deliberately left, both worth their own issues:

1. The wedge's `role_matches` models symmetry but has **no notion of `InverseObjectProperties` between two distinct roles** at all.
2. Default `classify` prints `fragment: Horn (… complete)` on the #128 file while missing a subsumption — `completeness_guaranteed()` returns true where it should not. That's a calibration bug independent of the chain rule.

## Notes for re-running the harnesses

These cost most of the debugging time and aren't written down anywhere:

- **`fetch-real-ontologies.sh` truncates the corpus silently.** `protege.stanford.edu` is unreachable, so it dies at `pizza` under `set -e` and skips `ro`, `go-basic` and `wine` queued behind it. A `|| continue` per source would localise the failure.
- **`bench-corpus/pizza.ofn` carries two `Import(...)` declarations** and ROBOT hangs resolving them — which looks exactly like a wedged oracle build. Convert from an import-stripped copy; rustdl never resolves imports either, so both sides then see the same axioms. Also: that file is the *ontostart* pizza and closes at 111/112, so the reference `pizza 499` belongs to a different fixture.
- **ORE paths are Mac-only.** The pool is at `/data/dumontier/ore-run` on the Linux host, not `~/data/ore-run`; `ore-fp-sweep.sh` and `ore-fragment-prepass.sh` hardcode `/Users/micheldumontier/...`.
- **Run ORE before/after PAIRED, never as two sweeps.** `RUSTDL_TEST_PAIR_MS` is a wall-clock per-pair budget, so a contended run does less work and reports more MISSED; two separately scheduled sweeps let scheduling noise masquerade as a code difference. `ore-fp-sweep.sh` drives one binary, so a before/after built from two of its runs is not a valid comparison. Even paired, 3 of 610 rows came back DIFF — all TIMEOUT-vs-completed flips at the 90 s cap, splitting 2:1 in opposite directions. Re-run serially, two agree exactly; the third (`ore_ont_9684`) times out on *both* binaries at a 900 s cap and yields no data — counted as missing, not as agreement. It also completed in under 90 s while contended and then failed to finish in 900 s uncontended: a wall-clock per-pair budget makes total runtime **non-monotone in available CPU**, because starved pairs blow their 200 ms deadline and get abandoned early. Timeout sets from these sweeps are not a stable property of an ontology.
- **Cap container *swap*, not just container memory.** Eight Konclude containers at `--memory=24g` took this host from 0 to 45 GB of swap (of 57) at ~55 MB/s sustained swap-out: a cgroup memory limit does not prevent host paging, it *causes* reclaim, which the kernel satisfies by swapping. Set `--memory-swap` equal to `--memory` so an oversized ontology OOMs in seconds and gets recorded.

Coverage stated as data, not as a pass: 463 of 628 ORE trigger ontologies (74%) compared; 146 hit the 90 s cap on at least one binary and 18 never got an oracle (Konclude OOM under a deliberate 12 GB no-swap cap, 552 MB down to 2.7 MB). Those 165 are missing data. Still unobtainable on this host and unchanged by this work: `galen`, `notgalen`, `alehif-test`, `ore-10908-sroiq`, `ore-15672-shoin`, `shoiq-knowledge`, `bibtex`, and the four `*-stripped` corpus entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,184 @@ All notable changes to rustdl are documented here. Format is based on
 
 ## [Unreleased]
 
+### Fixed — a role chain leg could not be walked through an inverse-equivalent edge (#128)
+
+`apply_role_chains` resolved each chain position against raw edge labels *and*
+directions: a `Named(r)` position read only `r`-labelled out-edges. So a chain whose
+first leg had to be traversed **backwards** never fired, even when an
+`InverseObjectProperties` declaration made that hop an identity rather than an
+assumption. `SymmetricObjectProperty(r)` lowers to `InverseObjectProperties(r, r)`, so
+symmetric roles were the common way to hit it — the reported shape was
+
+```
+Symmetric(coOccursWith),  coOccursWith ∘ precedes ⊑ precedes
+DiscomfortT1a ≡ Discomfort ⊓ ∃coOccursWith.Yawning ⊓ ∃precedes.Pain
+DiscomfortT1b ≡ Discomfort ⊓ ∃coOccursWith.(Yawning ⊓ ∃precedes.Pain)
+```
+
+where `HermiT` reports the two classes equivalent and rustdl reported only
+`T1b ⊑ T1a` — the direction that reads the chain forwards. `T1a ⊑ T1b` needs the chain
+to fire at the `Yawning` witness `y` through `coOccursWith(y, x)`, which exists purely
+by symmetry.
+
+Chain legs now accept an opposite-direction edge whose role is a declared inverse of
+the wanted one, via the same `are_declared_inverses` predicate `edge_satisfies` already
+used for the cross-polarity concept-level match — so the chain rule and the `∃`/`∀`
+matcher no longer disagree about which edges exist. `r ≡ s⁻` is an equivalence, so the
+extra hop derives nothing a model could refuse.
+
+**Both reads compose with the role hierarchy**, because nothing materialises super-role
+edges: `apply_role_rules` adds concept *labels* and `add_edge` records only the exact
+role, the rest of the tableau consulting the hierarchy lazily via `edge_satisfies`. So
+forward accepts any `q ⊑ wanted`, and reverse accepts a `q` reaching some super-role
+that is a declared inverse of `wanted`. Load-bearing, not theoretical: with `q ⊑ co`,
+`Symmetric(co)` and `co ∘ p ⊑ p`, deriving `∃q.Y ⊓ ∃p.Pn ⊑ ∃q.(Y ⊓ ∃p.Pn)` needs
+`q(x,y) ⇒ co(x,y) ⇒ co(y,x)` before the chain can fire, and no EL closure can rescue it
+because symmetry is out of fragment. Every accepted edge is licensed by a `⊑` or a `≡`;
+`has_inverse_pairs()` short-circuits the whole reverse path on an inverse-free ontology.
+
+> **RETRACTED CLAIM.** An earlier revision of this entry said the two negative tests
+> were "FP guards … that a blanket *read in-edges too* fix would trip". **They were
+> not.** Both fixtures declared no symmetry and no inverse pair, which makes them pure
+> EL: `is_subclass_of` returns the saturation-closure verdict and `chain_leg_targets`
+> never executes. Mutating the reverse read to `true || are_declared_inverses(..)` — the
+> exact unsound fix they were said to catch — left **all six tests passing**. Found by
+> an adversarial review, not by the suite.
+>
+> Two lessons are now written into the test file. (1) **Every tableau fixture must be
+> out of the EL fragment**, or it pins the saturator instead of the code under test —
+> and an EL-derivable *positive* is masked the same way, which is why a chain-plus-
+> sub-role "control" cannot demonstrate anything about the chain rule. (2) Leaving EL is
+> harder than it looks: `BareRoleDecls` admits an inverse/symmetry declaration over a
+> role no axiom can *read* as semantically inert, so a pair over unused roles keeps the
+> ontology pure EL. The declaration has to be over an **observable** role.
+>
+> The suite is now 8 tests, all non-EL, and the file records the acceptance criterion:
+> **a test that survives that mutation is not a guard.** Both negatives now fail under
+> it, and under a second mutation that walks `super_roles` without checking the inverse
+> partner.
+
+`is_subclass_of` now answers the reporter's ontology correctly in both directions.
+**Default `classify` still does not** — it never reaches the chain rule on this shape,
+because the hypertableau wedge decides the pair first and the wedge's `role_matches`
+can traverse a symmetric edge backwards only when the role hierarchy is threaded in,
+which happens solely under `RUSTDL_CLASSIFY_SAME_TIER=1` (documented "DEFAULT OFF —
+sound (FP=0) but corpus-invisible and ~2× wall"). #128 is the counterexample to
+"corpus-invisible", so that default is now a live question rather than a settled one.
+The remaining gap is pinned — visibly, not `#[ignore]`d — by
+`default_classify_still_misses_the_symmetric_chain_leg`, and the wedge additionally has
+no notion of `InverseObjectProperties` between two *distinct* roles at all.
+
+Closure-identical on every ontology available offline (94 files: `bench-corpus`, the
+`docs` reduced cores, and the 87 bench fixtures — 0 changed), and the 87-fixture
+ROBOT/HermiT-oracle differential suite still passes.
+
+**Corpus soundness net: run, and byte-identical to a pristine-HEAD baseline.**
+`./scripts/run-soundness-diff.sh` was run twice over freshly provisioned fixtures —
+once on this change, once on a `HEAD` worktree sharing the same `ontologies/real` — and
+the `[fp0]` manifests, the per-subsumption MISSED/FP lists, and the failure sets all
+diff clean. `ro 158=158`, `sio 8904=8904`, `sulo 51=51`, `wine 653=653`, all
+`FP=0 MISSED=0`, and `family` inconsistency detected. Those four exactly reproduce the
+reference values recorded in `CLAUDE.md` from oracles generated here by a different
+Konclude build, which is the pipeline-validation control that makes the rest of the run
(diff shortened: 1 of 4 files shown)


micheldumontier:
Cross-references, spotted after opening this:

- **#131** (range on a chain's second leg or super role) is the other mechanism in the #128 file. Independent of this PR, but worth noting that with this fix the reporter's ontology clears **without** the `Range: Process` removal #131's investigation needed — so #131 is a real separate defect that simply isn't what gates that entailment once the symmetry hop works. Both live in the chain rule's neighbourhood; whoever takes #131 should read `chain_leg_targets` here first.
- **#124** already covers the false `completeness_guaranteed()` certificate that the *What this does NOT fix* section describes. I filed a duplicate (#134) before spotting it and have closed that in favour of #124, adding the #128 counterexample there.
- **#135** tracks the wedge-side gap this PR deliberately leaves.
- **#130** already corrects the stale ORE path in `CLAUDE.md`, so that bullet in the harness notes above overlaps it. The other four notes (fetch-script truncation under `set -e`, pizza's unresolvable imports, paired-vs-two-sweeps, and capping container *swap* rather than just memory) are not recorded anywhere yet and are the ones that actually cost time.

micheldumontier:
Rebased onto `main` (`6bb6cc2`, post horned-owl 3.0). The test file's `read_ofn`/`SetOntology<RcStr>`/`ParserConfiguration` imports are unchanged by the 3.0 bump, so no source changes were needed; all 6 tests still pass, fmt and clippy clean.

**Two things a reviewer should know before running the full suite locally.**

1. `cargo test` fails on `owl-dl-cli --test incremental_fixpoint_identity` **if you have `ontologies/real/` provisioned**. That is not this PR — it reproduces identically on pristine `main`, and I filed it as #137: `find_open_at_most`'s `≤1` invariant is false on pizza, so a debug build panics on the default config while release silently skips the violation. The test only passes today because the fixture is gitignored and absent. CI is green here for the same reason.

2. The ORE and corpus numbers in the PR body were measured **pre-horned-owl-3.0**. #136 records a 3.0 parse regression on 7 ORE ontologies, so a re-run on this base would not be directly comparable. I have deliberately not refreshed them rather than silently mixing two dependency versions in one table — say if you would rather I re-measure on the new base and label both.

Also: @PiotrNowara posted a second reproducer on #128 that this PR does **not** fix, now split out as #138 (an inverse edge failing to discharge an `∃` on a chain-reached successor). I localised it against this branch — the chain-plus-sub-role control passes, so it is a distinct mechanism from the one fixed here, not a gap in this fix.

PR #140, opened 2026-09-10
fix(hyper): the incremental Horn drain silently lost subsumptions (#137)

### LINKED ISSUE #137: find_open_at_most's ≤1 invariant is false on pizza: debug panics on the default config, release silently skips the violation
## Summary

`hyper.rs`'s `find_open_at_most` asserts an invariant that **does not hold** on a real corpus ontology. On a debug build this panics (`rc=101`) on the **default** configuration; on release the `debug_assert!` compiles out and the code `continue`s, silently skipping a genuine `≤1` violation.

Found while provisioning `ontologies/real/` for #133's soundness run. **Pre-existing and unrelated to that PR** — reproduced identically on pristine `main` (`6bb6cc2`).

## Reproducer

```console
$ ./target/debug/rustdl classify ontologies/real/pizza.ofn
thread 'main' panicked at crates/owl-dl-tableau/src/hyper.rs:3871:21:
≤1 violation reached find_open_at_most under inverse_func_merge
$ echo $?
101
```

`RUSTDL_HYPER_INCREMENTAL_FIXPOINT` defaults to ON (`is_none_or(|v| v != "0" && !v.is_empty())`), so **no flag is needed to hit this**. Fully deterministic — 3/3 runs, and it persists with `--pair-timeout-ms 0`, so it is not the per-pair budget.

| build | flag | verdict lines | outcome |
|---|---|---|---|
| debug | ON (default) | 0 | **panic, rc=101** |
| debug | `=0` | 69 | ok |
| release | ON (default) | 69 | ok |
| release | `=0` | 69 | ok |

The fixture is `bench-corpus/pizza.ofn` (md5 `8be586bc…`) copied to `ontologies/real/pizza.ofn`; the ontostart pizza, 92 classes. Note the debug/release split — the panic is a `debug_assert!`, so release does not crash but takes the same skipping branch.

## The invariant, and why release is the worrying case

```rust
if self.inverse_func_merge && n == 1 {
    // Deterministic ≤1/functional merges fire incrementally in
    // `horn_fixpoint` (both the edge and constraint-add
    // triggers), so no `n == 1` violation should survive to the
    // branch layer; only genuine nondeterministic `n ≥ 2` does.
    debug_assert!(
        self.distinct_role_succ(node, role, qual).len() <= 1,
        "≤1 violation reached find_open_at_most under inverse_func_merge"
    );
    continue;
}
```

The `continue` is **unconditional** — it runs whether or not the assertion holds. So in release, a node that genuinely has ≥2 distinct successors under a `≤1` constraint is skipped by the branch layer, and the merge that would expose a clash never happens.

Direction of the resulting error: a missed merge means a missed clash, so a branch is reported satisfiable when it is not — i.e. a **MISSED subsumption, not a false positive**. Sound but incomplete, and silent. That matches the D10 pattern this project already tracks (wrong answer, no signal): `classify` on this ontology reports `# label heuristic: pruned=2690 pass_through=2 misses=0` with no indication that a constraint violation was discarded.

Worth deciding which of these is true, because the fix differs:

1. The invariant is correct and the incremental drain has a hole — `horn_fixpoint`'s `≤1` triggers are not firing for some edge/constraint arrival order. Then the fix is in the drain, and the assert stays.
2. The invariant is too strong — some `≤1` violation legitimately reaches the branch layer. Then the `continue` is a real bug: it must fall through to the merge path for `n == 1` as well, and the comment is wrong.

The flag-OFF path getting the right answer (69 lines both builds) points at (1): with the non-incremental re-seed, the `≤1` merge does fire.

## Second-order finding: the gate was vacuous

`crates/owl-dl-cli/tests/incremental_fixpoint_identity.rs` — the "durable byte-identity gate" on this exact flag — **fails the moment the corpus is present**:

```
assertion `left == right` failed: mismatch on ontologies/real/pizza.ofn
  left: [70 direct subsumptions]   right: []
```

It has been passing only because `ontologies/real/pizza.ofn` is gitignored and absent in CI and in a fresh checkout, and the test skips absent `ontologies/` fixtures by design. Its `compared >= 2` guard is satisfied by the two checked-in bench fixtures, so it reports `ok` while never exercising the one fixture that catches this. Same shape as the `[fp0]` net note in `CLAUDE.md` ("of 22 fixture blocks, 9 could skip while the suite still reported `ok`").

Suggest either checking in a small fixture that reproduces this, or making the skip loud (print a `SKIPPED` line per absent fixture) so a vacuous run is visible.

Related: #123 (panics instead of erroring on real ontologies) is the same failure *mode* on a different code path.

micheldumontier:
Fixes #137, and closes the corpus net's last MISSED.

## The bug

`RUSTDL_HYPER_INCREMENTAL_FIXPOINT` (**default ON**) drains only the delta a decision pushed. That is correct only if every rule has a trigger for every way its premises can arrive. The non-incremental path re-seeds from the graph every pass and so *masks* a missing trigger; incremental mode simply never fires the rule and reports `Sat` — and with `trust_sat` on (also default) that spurious `Sat` is believed, the pair never reaches the tableau, and the subsumption is lost **with no signal**.

Surfaced by provisioning `ontologies/real/` for #133's soundness run, which activated a test that had been quietly skipping.

## Three holes

1. **Qualified-`≤1` label trigger.** `≤1 r.C` counts only the `r`-successors carrying `C`, and the label can arrive *after* the edge, with no `Event::Edge` to notice. Every violation I dumped was `qual=Some(..)`, never unqualified — which is exactly why functional-only corpora (galen, notgalen) never showed this. Fixed with `at_most_ones_qualified` plus a neighbour walk in the `Event::Label` arm, mirroring the edge trigger's src/tgt symmetry.
2. **Merge-inherited `≤1`.** A merge copies `s_j.at_most` onto the survivor without firing the constraint-added trigger that the `Atom::AtMost` head path gets, so a survivor already holding ≥2 matching successors was never re-checked.
3. **The general case**, because (1) and (2) are unlikely to be the last: on reaching quiescence, re-seed once and re-drain. Sound by construction — it can only add entailed facts, i.e. only turn a wrong `Sat` into `Unsat`. Gated by `RUSTDL_RESEED_VERIFY` (**default ON**, `=0` reverts).

**The net is load-bearing 142 times on `sio` and twice on `ro`**, so the drain's incompleteness is widespread — it just happened to change a verdict only on pizza. So a residual hole remains: `stats.reseed_recovered` counts it and `RUSTDL_RESEED_PROBE=1` prints each occurrence. That is deliberately a **counter, not a `debug_assert`** — asserting "no trigger is ever missing" fires immediately, because it is precisely what is still false. Locating the last trigger is follow-up work; the verdict is correct either way.

## Two recoveries, both oracle-confirmed

- **`pizza`: 111/112 → 112/112, FP=0, MISSED=0.** `SpicySalamiPizza ⊑ SpicyPizza`, which HermiT and Konclude both derive, was being dropped on the default config. **Every curated-corpus fixture is now MISSED=0.**
- **A pinned known-miss now derives.** `dkey_collapse_broadcast`'s `KNOWN-MISS-forall-super-value-sub` (`∀f.DataOneOf("a")`, `p ⊑ f`, `p(i,"b")`) is logically inconsistent and rustdl reported it consistent. Its own comment said to flip the expectation once the gap closed — done. Note the asymmetry that makes this a #137 fix rather than a datatype one: `RUSTDL_HYPER_INCREMENTAL_FIXPOINT=0` still misses it.

## Cost

A/B-able via the flag: `sio` 1.07 → 1.15 s (+7%), `ro` 0.30 → 0.60 s, `pizza` and `wine` unchanged.

## The gate was vacuous — the transferable part

`incremental_fixpoint_identity.rs` is described in its own header as the "durable byte-identity gate" on this flag. The only fixture in it that ever caught a real bug was the **gitignored** `ontologies/real/pizza.ofn` — absent in CI and in every fresh checkout — while its `compared >= 2` guard was satisfied by two checked-in bench fixtures. So it reported `ok` for as long as nobody provisioned the corpus by hand, and the default config silently lost a subsumption the whole time.

That pizza pair is now reduced to a 7-class fixture and **checked in**. Verified by moving pizza aside (simulating CI): the gate fails on the new fixture under `RUSTDL_RESEED_VERIFY=0` and passes with the net on. Same shape as the `[fp0]` net's recorded "9 of 22 fixture blocks could skip while the suite still reported `ok`" — the other gitignored-fixture gates are worth the same audit.

## Verification

Full suite + doctests green (including the previously-failing identity gate); fmt and clippy `-D warnings` clean; corpus soundness net with every fixture FP=0/MISSED=0 — `sio` 8904, `wine` 653, `ro` 158, `sulo` 51, `pizza` 112, `family` inconsistency detected. The net's 7 failures are the unchanged fixture-absence panics for `galen`/`notgalen`/`alehif`/ORE/`bibtex`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,66 @@ now states outright that **`incomplete: false` does not mean complete**, and tha
 `disjoint --json` reads `completeness_guaranteed()` directly, so it now correctly reports
 `incomplete: true` on this ontology — which was #124's headline symptom.
 
+### Fixed — the incremental Horn drain silently lost subsumptions; three missing triggers (#137)
+
+`RUSTDL_HYPER_INCREMENTAL_FIXPOINT` (**default ON**) drains only the delta a decision
+pushed, which is correct only if every rule has a trigger for every way its premises can
+arrive. The non-incremental path re-seeds from the graph every pass and so masks a
+missing trigger; incremental mode simply never fires the rule and reports `Sat` — and
+with `trust_sat` on (also default) that spurious `Sat` is believed, the pair never
+reaches the tableau, and the subsumption is lost **with no signal**.
+
+Found by provisioning `ontologies/real/` for #133's soundness run, which activated a
+test that had been skipping. Three distinct holes, all fixed:
+
+1. **The qualified-`≤1` label trigger.** `≤1 r.C` counts only the `r`-successors
+   carrying `C`, and the label can arrive *after* the edge, with no `Event::Edge` to
+   notice. Every violation observed was `qual=Some(..)`, never unqualified — which is
+   why functional-only corpora (galen, notgalen) never showed it. Fixed by
+   `at_most_ones_qualified` + a neighbour walk in the `Event::Label` arm, mirroring the
+   edge trigger's src/tgt symmetry.
+2. **Merge-inherited `≤1`.** A merge copies `s_j`'s `at_most` onto the survivor without
+   firing the constraint-added trigger the `Atom::AtMost` head path gets, so a survivor
+   already holding ≥2 matching successors was never re-checked.
+3. **The general case**, since (1) and (2) are unlikely to be the last: a completeness
+   net. On reaching quiescence `horn_fixpoint` re-seeds once and re-drains, so a rule
+   whose premise arrived without a trigger still fires. Sound by construction — it can
+   only add entailed facts, i.e. only turn a wrong `Sat` into `Unsat`. Gated by
+   `RUSTDL_RESEED_VERIFY` (**default ON**, `=0` reverts).
+
+**The net is load-bearing 142 times on `sio` and twice on `ro`**, so the drain's
+incompleteness is widespread; it happened to change a verdict only on pizza. A residual
+hole therefore remains — `stats.reseed_recovered` counts it and `RUSTDL_RESEED_PROBE=1`
+prints each occurrence. It is deliberately a COUNTER, not a `debug_assert`: asserting
+"no trigger is ever missing" fires immediately, because that is exactly what is still
+false. Finding the last trigger is follow-up work; the verdict is correct either way.
+
+**Two completeness recoveries, both oracle-confirmed:**
+
+- **`pizza` closes the corpus net's last MISSED: 111/112 → 112/112, FP=0, MISSED=0.**
+  `SpicySalamiPizza ⊑ SpicyPizza` (HermiT and Konclude both derive it) was being dropped
+  on the default config. Every curated-corpus fixture is now MISSED=0.
+- **A pinned known-miss now derives.** `dkey_collapse_broadcast`'s
+  `KNOWN-MISS-forall-super-value-sub` — `∀f.DataOneOf("a")`, `p ⊑ f`, `p(i,"b")` — is
+  logically inconsistent and rustdl reported it consistent. Its own comment said to flip
+  the expectation once the gap closed; done. Note the asymmetry that makes this a #137
+  fix and not a data-property one: `RUSTDL_HYPER_INCREMENTAL_FIXPOINT=0` still misses it.
+
+Cost, A/B-able via the flag: `sio` 1.07 → 1.15 s (+7%), `ro` 0.30 → 0.60 s, `pizza` and
+`wine` unchanged.
+
+**The gate that should have caught this was passing vacuously**, which is the more
+transferable lesson. `incremental_fixpoint_identity.rs` is described as the "durable
+byte-identity gate" on this flag, and its only fixture that ever caught a real bug was
+the **gitignored** `ontologies/real/pizza.ofn` — absent in CI and in every fresh
+checkout, while its `compared >= 2` guard was satisfied by two checked-in bench fixtures.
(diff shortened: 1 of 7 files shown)


micheldumontier:
**This also fixes #131**, which turns out to be the same defect rather than a separate range/chain gap.

Verified by toggling this PR's net alone on #131's reproducer:

| | `equiv(A,B)` |
|---|---|
| this branch | **1** ✓ |
| `RUSTDL_RESEED_VERIFY=0` | **0** |

#131's own `effective_ranges`/`chain_ranges` hypothesis is ruled out: `ObjectPropertyRange` clausifies to a plain Horn rule, `classify --saturation-only` is identical with and without the range, and `subclass` is correct in both cases. The range does not break a rule — it *creates* a derivation the delta drain fails to reach. Details on #131.

Consequence worth noting for merge order: **#128's reporter gets a correct `classify` on their original file only with this PR AND #142** (the wedge-hierarchy fix). Verified on a local merge of the two — `equiv DiscomfortT1a DiscomfortT1b`, unmodified file, no flags. Neither PR alone is sufficient: this one alone still gives 0 on their file, because the wedge cannot see symmetry without #142.

So #131 can be closed by this PR, and #128 by this plus #142.

Issue #135, opened 2026-09-08
hypertableau role_matches has no notion of InverseObjectProperties between distinct roles (#128)


micheldumontier:
Split out of #128. The classic tableau's chain rule was fixed there (#133); the hypertableau has its own, narrower version of the same gap, and it is the engine that decides pairs on the default `classify` path.

## What the wedge does and does not model

`hyper::role_matches` handles **symmetry**:

```rust
if let Some(h) = sub_roles
    && edge.role_id() == wanted.role_id()
    && h.is_symmetric(wanted.role_id())
{
    return true;
}
```

That test requires `edge.role_id() == wanted.role_id()`, so it only ever relates a role to *itself*. There is no `are_declared_inverses` equivalent anywhere in `hyper.rs` — `InverseObjectProperties(r, s)` between two **distinct** roles is invisible to the wedge's body matching.

`TableauContext::edge_satisfies` (classic tableau) does model it, via `inverse_pairs_set`. So the two engines disagree about which edges exist, in the same way #133 fixed the chain rule disagreeing with `edge_satisfies`.

## Repro

```
ObjectProperty: r  InverseOf: s
ObjectProperty: p  SubPropertyChain: r o p
Class: A EquivalentTo: (s some Y) and (p some Pain)
Class: B EquivalentTo: (s some (Y and (p some Pain)))
```

`A ⊑ B` is entailed: `s(x,y)` *is* `r(y,x)`, so `r ∘ p ⊑ p` fires at `y` and gives `p(y,z)`. Post-#133 `is_subclass_of` gets this right; the wedge does not, and `RUSTDL_CLASSIFY_SAME_TIER=1` does **not** rescue it (that flag supplies the hierarchy, which carries sub-roles and the symmetric flag — not inverse pairs). Pinned as `a_declared_inverse_pair_fires_the_chain_at_the_successor` in `crates/owl-dl-reasoner/tests/chain_symmetric_leg.rs`, which queries `is_subclass_of` for exactly this reason.

## Two candidate directions

1. **Give the wedge the pairs.** Thread `inverse_pairs` into `HyperEngine` alongside `sub_roles` and extend `role_matches` with the cross-polarity `are_declared_inverses` arm, mirroring `edge_satisfies`. `role_atom_holds` must move in lockstep — its own doc comment already warns that any divergence from `enumerate_matches` is "a completeness or FP bug depending on its direction".
[...]

PR #142, opened 2026-09-11
fix(classify): the wedge ran without the role hierarchy, hiding symmetry (#135)

### LINKED ISSUE #128: Missing expected equivalent classes (symmetric property with a property chain)

micheldumontier:
Refs #135, refs #128 — **partial**. Fixes the half that makes symmetry visible to `classify`; the distinct-role inverse-pair half is not fixed and is explained below.

## The bug

`RUSTDL_CLASSIFY_SAME_TIER` bundled two independent things behind one **default-OFF** lever, justified by a "~2× wall":

- **Layer A** — carry the role hierarchy into the classify wedge
- **Layer B** — broaden the same-tier sweep to label-driven

With Layer A off, every wedge on the default path ran with `sub_roles == None`, so `hyper::role_matches` fell back to exact role-id **and** exact polarity. **Symmetry and sub-role matching were invisible to the engine that decides most pairs.** That is why #128's entailment still vanished from `classify` after the calculus was fixed in #133 — `subclass` was right and `classify` wrong, on the same binary and the same pair.

Layer A is now its own lever, `RUSTDL_CLASSIFY_WEDGE_HIERARCHY` (**default ON**, `=0` reverts). Layer B keeps the old flag.

## Both halves of the old justification were wrong

**The "~2× wall" was Layer B's.** Layer A alone: `sio` 1.07 → 1.36 s, `wine` 3.17 → 3.99 s (~25-30%), `ro` 0.27 → 0.29 s, `pizza` unchanged.

**"Corpus-invisible" is true of the corpus and false of ORE — by 313 subsumptions.** Paired A/B over the 610-ontology trigger set, 463 with both binaries completing:

| metric | baseline | with Layer A |
| --- | --- | --- |
| rustdl subsumptions | 6,458,921 | 6,459,234 |
| oracle subsumptions | 6,462,873 | 6,462,873 |
| **false positives** | **0** | **0** |
| MISSED | 4,265 | **3,952** |

`ore_ont_16457` alone goes 936 → 695 MISSED; `ore_ont_16666`, `ore_ont_2792`, `ore_ont_8148` each recover ~20. Of the 13 DIFF rows, 4 are TIMEOUT flips at the 90 s cap in **both** directions, and `ore_ont_9662` looked like a regression (MISSED 1 → 2) but is budget noise — identical at a 600 s cap, 3/3 runs. I re-ran that row specifically because it was the only one pointing the wrong way.

Corpus soundness net unchanged, every fixture FP=0.

## This does NOT close #128 on its own

The reporter's ontology needs a third fix:

| `epi_corrected_debug_minimal` | default | with Layer A |
| --- | --- | --- |
| as submitted | 0 | **0** |
| `Domain`/`Range` stripped | 0 | **1** ✓ |

So **#131 co-blocks it**. #128 needs #133 (merged) + this + #131. Leaving #128 open.

## #135's other half is NOT fixed

`role_matches` still has no notion of `InverseObjectProperties` between two **distinct** roles; only the `r == s` (symmetry) case works. Three attempts failed:

1. carrying the pairs on `RoleHierarchy`,
2. a cross-polarity arm in `role_matches`,
3. indexing the first-leg trigger under the partner's key (the `Event::Edge` lookup is keyed by the **edge's** role, which is why symmetry gets away with the atom's own key and a distinct partner does not).

All three were reverted rather than left as dead code. The reassessment worth recording: the codebase already has a strategy for distinct inverse pairs — **canonicalisation**, `build_role_hierarchy`'s `canon_map` rewriting `s → co⁻` — and teaching a second matcher about a second representation fights it. The route worth trying is applying `canon` during clausification so the case collapses into the polarity handling the wedge already has. That is a clausifier change and deserves its own attempt, not a fourth patch on three failures.

## Verification

fmt and clippy `-D warnings` clean; corpus soundness net all FP=0 with closures unchanged; ORE as above. Full suite green except `incremental_fixpoint_identity`, which is **#137** — it fails on `main` too whenever `ontologies/real/` is provisioned, and is fixed in **#140**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,59 @@ All notable changes to rustdl are documented here. Format is based on
 
 ## [Unreleased]
 
+### Fixed — the classify wedge ran without the role hierarchy, hiding symmetry (#135, #128)
+
+`RUSTDL_CLASSIFY_SAME_TIER` bundled two independent things behind one **default-OFF**
+lever, justified by a "~2× wall":
+
+- **Layer A** — carry the role hierarchy into the classify wedge.
+- **Layer B** — broaden the same-tier sweep to label-driven.
+
+With Layer A off, every wedge on the default path ran with `sub_roles == None`, so
+`hyper::role_matches` fell back to exact role-id **and** exact polarity: **symmetry and
+sub-role matching were invisible to the engine that decides most pairs**. That is why
+#128's symmetric-chain entailment still vanished from `classify` after the calculus was
+fixed in #133 — `subclass` was right and `classify` was wrong on the same binary.
+
+Layer A is now its own lever, `RUSTDL_CLASSIFY_WEDGE_HIERARCHY`, **default ON**
+(`=0` reverts). Layer B keeps `RUSTDL_CLASSIFY_SAME_TIER`.
+
+**The "~2× wall" belonged to Layer B.** Layer A alone, measured: `sio` 1.07 → 1.36 s,
+`wine` 3.17 → 3.99 s (~25-30%), `ro` 0.27 → 0.29 s, `pizza` unchanged.
+
+**And "corpus-invisible" was wrong — it is ORE-visible by 313 subsumptions.** Paired A/B
+over the 610-ontology trigger set, 463 with both binaries completing:
+
+| metric | baseline | with Layer A |
+| --- | --- | --- |
+| rustdl subsumptions | 6,458,921 | 6,459,234 |
+| false positives | **0** | **0** |
+| MISSED | 4,265 | **3,952** |
+
+`ore_ont_16457` alone goes 936 → 695 MISSED; `ore_ont_16666`, `ore_ont_2792` and
+`ore_ont_8148` each recover ~20. Of the 13 DIFF rows, 4 are TIMEOUT flips at the 90 s cap
+in both directions and one (`ore_ont_9662`) looked like a regression (MISSED 1 → 2) but is
+budget noise — identical at a 600 s cap, 3/3 runs. Corpus net unchanged, all FP=0.
+
+**This does NOT close #128 on its own.** The reporter's file needs a third fix: with
+Domain/Range stripped the entailment now appears in `classify`, with them present it does
+not — so #131 (range on a chain leg) co-blocks it. Measured:
+
+| `epi_corrected_debug_minimal` | default | with Layer A |
+| --- | --- | --- |
+| as submitted | 0 | 0 |
+| `Domain`/`Range` stripped | 0 | **1** |
+
+**#135's other half is NOT fixed.** `role_matches` still has no notion of
+`InverseObjectProperties` between two DISTINCT roles; only the `r == s` (symmetry) case
+works. Three attempts failed — adding the pairs to `RoleHierarchy`, a cross-polarity arm
+in `role_matches`, and indexing the first-leg trigger under the partner's key — and were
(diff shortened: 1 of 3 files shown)


micheldumontier:
Superseded by #143, which landed the same Layer A / Layer B split an hour earlier and with better evidence. Closing without merging.

**#143 is right to leave the flag default OFF and I was wrong to flip it.** Its doc records a full 1,920-ontology two-arm sweep — 1,800 identical, 23 gains, **2 regressed** — with both regressions reproducing sequentially on an idle host at a 300 s cap with identical answers in both arms: `ore_ont_16372` 3–4 s → 66–73 s (~19×) and `ore_ont_9890` 17–18 s → 89–94 s (~5×). The stated retry condition, *"a flip needs 0 REGRESSED"*, is not met.

My cost measurement was a **4-ontology sample** (sio, ro, wine, pizza) giving "~25-30%, not 2×". That is precisely the failure mode #143's doc calls out two paragraphs later: *"A 6-ONTOLOGY SAMPLE SAID THIS WAS FREE… The full sweep found 25. Sizing a default flip on a sample is the failure mode this repo already records."* I did the same thing with four.

**And I misread my own data.** `ore_ont_16372` appears in my paired ORE sweep as `base=0/0/FP=0/MISSED=0 → fixed=TIMEOUT`, which I classified as a TIMEOUT flip and dismissed as scheduling noise along with three others. It is one of the two real, reproducible wall regressions. I re-ran the single row that pointed the wrong way on *completeness* (`ore_ont_9662`) but did not re-run the timeout flips, so the one data point that contradicted my conclusion was the one I did not check.

## What is worth keeping from this branch

The **completeness** side of my sweep, which is complementary to #143's wall measurement and which I have not seen recorded there. Paired A/B over the 610-ontology trigger set (a chain leg meeting an inverse-equivalence on the same role), 463 with both binaries completing:

| metric | Layer A OFF | Layer A ON |
| --- | --- | --- |
| false positives | 0 | **0** |
| MISSED | 4,265 | **3,952** |

**313 subsumptions recovered, FP=0.** `ore_ont_16457` alone goes 936 → 695 MISSED; `ore_ont_16666`, `ore_ont_2792` and `ore_ont_8148` each recover ~20. So the flip is worth real completeness — which raises the value of pinning the wall cost rather than abandoning the question. That is the trade #143's retry conditions describe, now with a number on the benefit side.

Also worth noting for whoever takes the retry: **`ore_ont_16372` is a 0-row ontology** (`0/0`, 744 unsatisfiable classes) — it derives nothing either way, so its ~19× is pure overhead on a degenerate input and may be easier to pin than a general slowdown.

The test-pin flip on this branch is not needed either: #143 handled it, and `main` is green.

Issue #147, opened 2026-09-12
RUSTDL_SEMANTIC_BRANCHING=1 aborts with a rayon stack overflow on ore_ont_9890 (rc=134)


micheldumontier:
`RUSTDL_SEMANTIC_BRANCHING=1` aborts with a stack overflow in a rayon worker on a real ORE ontology. Default-off, so nobody hits it by accident — but it means the lever cannot be evaluated on the workload that most needs it.

## Reproducer

```console
$ RUSTDL_SEMANTIC_BRANCHING=1 ./target/release/rustdl classify ore_ont_9890.owl
thread '<unknown>' has overflowed its stack
Aborted (core dumped)      # rc=134
```

Deterministic — 2/2 runs, core dumped both times. The same ontology with the flag off completes normally (rc=0, ~17 s). `ore_ont_9890.owl` is `ore_ont_9890` from the ORE 2015 sample pool (0.3 MB, 627 classes).

An abort is worse than a slow answer: no partial result, no diagnostic, and `rc=134` rather than a `ReasonError`. Compare #123, which is the same failure *mode* (panic instead of erroring) on a different path.

## Why it matters beyond the flag

Found while profiling the `RUSTDL_CLASSIFY_ROLE_HIERARCHY` wall regression on #128. That regression is a **branch storm**: turning the role hierarchy on takes ⊔ branches from 126k to 543k on this exact ontology, and restores from 121k to 542k. Semantic branching's prune is precisely the shape that workload needs, so it is the most interesting lever to try there — and it is the one that crashes.

So this blocks an avenue on #128 rather than being purely cosmetic.

## Suspected shape

Stack overflow in a worker thread points at unbounded recursion rather than allocation pressure — `solve`/`horn_fixpoint` recursion depth under the semantic-branching path, with rayon's smaller default worker stacks making it surface there first. Worth checking whether the semantic-branching arm adds a recursive step that the depth cap (`HYPER_WEDGE_DEPTH` / iterative deepening) does not account for, since that cap is what bounds the ordinary path.

Not investigated further — filing so it is recorded with a reproducer.

micheldumontier:
Fixed on `main` by #154 (`8ba1b3f`). Verified on the merged commit: both previously-aborting configurations now complete with identical output, and forcing `RUSTDL_RAYON_STACK_MB=2` still reproduces the original abort — the control that the mechanism is understood rather than merely papered over.

[...]

PR #148, opened 2026-09-12
perf(classify): ask about inconsistency before paying for a hierarchy nobody reads (#128)

micheldumontier:
Refs #128. Removes **one of the two** wall regressions blocking the `RUSTDL_CLASSIFY_ROLE_HIERARCHY` default flip.

## The waste

`probe_says_inconsistent` runs **after** the tier walk. When it fires, `classify_inconsistent` discards the walk wholesale — every class is unsatisfiable, so the hierarchy just computed is binned. On `ore_ont_16372` that is **20.5 s of tier walk + 5.5 s of sweeps computed and thrown away**.

`RUSTDL_EARLY_INCONSISTENCY_PROBE` (**default OFF**, `=1` enables) asks first.

| `ore_ont_16372`, `RUSTDL_CLASSIFY_ROLE_HIERARCHY=1` | wall | rows |
| --- | --- | --- |
| early probe off | 31.6 s | 745 |
| early probe on | **5.9 s** | 745, byte-identical |

`ore_ont_16372` stops being a regression at all: the ON arm (5.9 s) is now **faster than the OFF arm** (9.1 s). `ore_ont_9890` is **not** fixed — its regression is genuine per-class wedge divergence under hierarchy-aware matching, not discarded work, and needs separate investigation before the flip.

## Two hidden gates

A straight hoist does nothing, and measuring why is the whole content of this change:

1. `unsatisfiable_idxs` is still **empty** before the walk — `ore_ont_16372`'s 744 unsat classes are discovered *by* the walk — so the probe's normal admission rejects.
2. There is a **second** `!incomplete_abox` bypass further down (the minimum-unsat-fraction guard, `classify.rs:1845`) which likewise rejects on a count that has not been computed yet.

Both are threaded by an explicit `extra_admission` parameter, passed `false` at the existing post-walk call site so its behaviour is bit-for-bit unchanged. I used a parameter rather than temporarily lending `stats.timed_out_pairs` a fake value, so a statistic is not repurposed to signal a callee.

## Admission and safety

The early call's evidence is a large `NoVerdict` share of the label cache — per-class wedge satisfiability that did not converge inside its budget. That is the same "did not look long enough" species the existing `RUSTDL_CLASSIFY_PROBE_ON_INCOMPLETE` arm already admits on, not a new idea.

A wrong admission costs **one bounded probe, never a wrong answer**: the probe *proves* inconsistency (layer 1 is an asserted instance of an unsatisfiable class; the rest are bounded wedge/⊤ probes under their own budget).

It **can** change output — on a KB where the late probe would never have been admitted but the early one is *and* the probe succeeds, the result becomes a proof-backed all-unsat where it was previously a partial hierarchy. That is a row **gain** at FP=0, not a loss, but it is a behaviour change, which is why this is default-OFF pending the two-arm sweep. A row-change sweep over the 610-ontology trigger set is running; results to follow.

## Corrections to the #128 record established here

- **The regression is a branch storm, not matcher cross-product.** Enumeration frames per match call are ≈1 in *both* arms — the hot bodies have 0–1 role atoms and essentially never recurse, so there is nothing to prune. Meanwhile ⊔ branches go 54k → 632k and restores 45k → 631k. Per-branch matcher cost *falls*; the search tree grows ~12×. My earlier "candidates multiply rather than add" explanation on #128 is withdrawn.
- **The all-zero phase banner is a reporting artifact.** `classify_inconsistent` rebuilds stats as `{ inconsistent, fragment, ..Default::default() }`, zeroing every timer *after* the work is done. The flag doc's "every classify phase reports 0 ms … the cost is not attributable from the banner" has a cause, and it is this.

Credit: the counter instrumentation and the original prototype measurement came from an independent analysis pass; I verified the stats-discard and the timing independently and reimplemented the fix.

## Verification

fmt, clippy `-D warnings`, full test suite green. Corpus soundness net every fixture FP=0/MISSED=0 — sio 8904, wine 653, ro 158, sulo 51, pizza 112.

Also filed **#147** on the way: `RUSTDL_SEMANTIC_BRANCHING=1` aborts with a rayon stack overflow on `ore_ont_9890`, which matters because that lever's prune is exactly the shape the remaining branch storm needs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,53 @@ All notable changes to rustdl are documented here. Format is based on
 
 ## [Unreleased]
 
+### Added — early inconsistency exit, removing one of the two blockers on the role-hierarchy flip (#128)
+
+`probe_says_inconsistent` runs AFTER the tier walk. When it fires,
+`classify_inconsistent` discards the walk wholesale — every class is unsatisfiable, so
+the hierarchy it just computed is thrown away. On `ore_ont_16372` that is **20.5 s of
+tier walk plus 5.5 s of sweeps computed and binned**. `RUSTDL_EARLY_INCONSISTENCY_PROBE`
+(**default OFF**, `=1` enables) asks first.
+
+| `ore_ont_16372`, `RUSTDL_CLASSIFY_ROLE_HIERARCHY=1` | wall | rows |
+| --- | --- | --- |
+| early probe off | 31.6 s | 745 |
+| early probe on | **5.9 s** | 745 (byte-identical) |
+
+That is one of the **two** ontologies whose wall regression keeps
+`RUSTDL_CLASSIFY_ROLE_HIERARCHY` (worth 313 recovered subsumptions at FP=0) default-OFF
+against a "0 REGRESSED" bar — and with this it stops being a regression at all: the ON
+arm (5.9 s) is now FASTER than the OFF arm (9.1 s). `ore_ont_9890` is **not** fixed and
+is the harder case; its regression is genuine per-class wedge divergence rather than
+discarded work.
+
+**Two hidden gates, both found by measurement after a straight hoist did nothing.**
+`unsatisfiable_idxs` is still EMPTY before the walk — those 744 unsat classes are found
+BY the walk — so the probe's normal admission rejects. And there is a SECOND
+`!incomplete_abox` bypass further down (the minimum-unsat-fraction guard) that likewise
+rejects on a count not yet computed. Both are now threaded by an explicit
+`extra_admission` parameter, `false` at the existing post-walk call site so its behaviour
+is unchanged.
+
+The early call's admission evidence is a large `NoVerdict` share of the label cache —
+per-class wedge satisfiability that did not converge in its budget, the same "did not
+look long enough" species the existing `RUSTDL_CLASSIFY_PROBE_ON_INCOMPLETE` arm already
+admits on. A wrong admission costs one bounded probe, never a wrong answer: the probe
+PROVES inconsistency. It CAN change output — a proof-backed all-unsat where the late
+probe was never admitted — which is a row gain at FP=0, and is why this ships default-OFF
+pending the two-arm sweep.
+
+**Two corrections to the #128 record while establishing this.** The regression is a
+BRANCH STORM, not matcher cross-product: enumeration frames per match call are ≈1 in both
+arms (nothing to prune), while ⊔ branches go 54k → 632k and restores 45k → 631k.
+Per-branch matcher cost actually falls; the tree grows. And the all-zero phase banner that
+made the cost unattributable is an artifact — `classify_inconsistent` rebuilds stats as
+`{ inconsistent, fragment, ..Default::default() }`, zeroing every timer after the work is
+done.
+
+Verification: fmt, clippy `-D warnings`, full test suite green; corpus soundness net every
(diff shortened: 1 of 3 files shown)


micheldumontier:
**Do not merge yet — my "can only gain rows" claim above is not supported by the sweep, and the sweep itself turned out to be the wrong instrument.**

The row-change sweep over the 610-ontology trigger set came back 591 SAME / 12 ROWS_GAINED / **6 ROWS_LOST** / 1 ROWS_CHANGED. A row loss on a KB that stays consistent would contradict the safety argument outright, so I chased all six.

They do not reproduce:

- Three (`ore_ont_12698`, `ore_ont_8911`, `ore_ont_9577`) are **nondeterministic at identical config** — same flags, two runs, different row counts. The "loss" is noise.
- `ore_ont_1707`: 3 paired runs, **3/3 identical**.
- `ore_ont_16457`: the ON arm produces *more* rows in 2 of 3 paired runs — a gain, not a loss.
- `ore_ont_1938` reported `rows=69840/0` because the ON arm hit the 60 s sweep cap, not because rows vanished.

So the honest conclusion is not "the six were fine" but: **at the default 5 ms per-pair budget the row-count signal is below the noise floor, and a sweep built on raw row deltas cannot answer this question.** That budget is wall-clock, so results move with machine load. I built the wrong instrument — the same mistake as sizing a default flip on a 4-ontology sample earlier in #128.

Re-running the flagged ontologies against the **Konclude oracles** instead, comparing FP/MISSED, which is what soundness actually rests on and is robust to a row appearing or vanishing at the margin. Will report that verdict here.

The `ore_ont_16372` result (31.6 s → 5.9 s, byte-identical) is unaffected by any of this — it is deterministic and was verified independently twice.

micheldumontier:
**Oracle verdict: clean. 18 SAME / 1 DIFF, FP=0 in every arm.**

Re-ran every ontology the row sweep had flagged, comparing closure size, FP and MISSED against the Konclude oracles instead of raw row counts:

| verdict | count |
|---|---|
| identical closure, FP and MISSED | **18** |
| DIFF | 1 |

The single DIFF is the target ontology:

```
ore_ont_16372   off=TIMEOUT   on=0/0/FP=0/MISSED=0
```

The OFF arm exhausts a 300 s cap; the ON arm completes. That is a pure gain, and the direction the change exists to produce.

**No FP in either arm on any ontology.** So the six apparent `ROWS_LOST` were entirely an artifact of measuring row counts at a wall-clock-sensitive 5 ms per-pair budget — three of them are nondeterministic at *identical* config, one is 3/3 identical under paired runs, one actually gains a row, and one was a sweep-cap DNF misread as a loss.

## What I got wrong here, for the record

My original safety claim — "it can only turn a partial hierarchy into a proof-backed all-unsat, so rows can only be gained" — was not *wrong* about the mechanism, but I had no evidence for it when I wrote it, and the instrument I built to check it could not have told me either way. Row deltas at that budget are below the noise floor. The oracle comparison is the instrument that answers the question, and it is the one the repo already uses for exactly this purpose.

Marking ready for review again. Unchanged and independently re-verified: `ore_ont_16372` 31.6 s → 5.9 s with byte-identical output, taking it from a wall regression to *faster than the OFF arm*.

Still default-OFF, and `ore_ont_9890` is still unfixed — so this removes one of the two blockers on the `RUSTDL_CLASSIFY_ROLE_HIERARCHY` flip, not both.

PR #151, opened 2026-09-12
fix(convert): read InverseObjectProperties with its polarity (#145)

### LINKED ISSUE #145: LATENT FP: collect_inverse_pairs strips polarity and mark_symmetric fires on matching ids — goes live if the horned-owl parser gap closes (blocks-after #126)
Found while reviewing #128. **Not reachable today.** Filing it now because what makes it unreachable is a *parser* limitation, and #126 (horned-owl 1.4 → 3.0) changes the parser.

## The hazard

Two places treat an `InverseObjectProperties` axiom by **role id**, discarding polarity:

- `collect_inverse_pairs` (`crates/owl-dl-reasoner/src/lib.rs`) pushes `(a.role_id(), b.role_id())`.
- `mark_symmetric` marks a role symmetric whenever the two ids in an `InverseObjectProperties` axiom match.

So, if these could ever be written:

| axiom | means | would be recorded as |
|---|---|---|
| `InverseObjectProperties(:p ObjectInverseOf(:p))` | a **tautology** (`p⁻` is `p`'s inverse) | `p` is **symmetric** — FALSE |
| `InverseObjectProperties(:p ObjectInverseOf(:q))` | `p ≡ q` | `p` and `q` are **inverses** — FALSE |

Both would be **false positives** — they assert role relationships the ontology does not state, and unlike a missed entailment an FP is unsound.

## Why it is unreachable today

horned-owl rejects `ObjectInverseOf` inside `InverseObjectProperties`:

```
Error: parsing OFN ontology …: Parsing Error:  --> 4:28
  |
4 | InverseObjectProperties(:p ObjectInverseOf(:p))
  |                            ^---
```

**Both Konclude and HermiT accept it**, so this is a rustdl-side parser gap, not invalid OWL. That gap is the only thing standing between the code above and a live unsoundness.

## Why now

#126 upgrades horned-owl 1.4 → 3.0. If the newer parser accepts the construct — deliberately or incidentally — this becomes reachable **without anyone touching the reasoner**, and it would surface as an unexplained FP in an area where nothing changed.

The guard is cheap and worth having in place *before* that lands, independent of whether 3.0 happens to accept it:

- `collect_inverse_pairs` should either respect polarity or explicitly reject same-id pairs rather than recording them as inverses.
- `mark_symmetric` should require **matching polarity**, not merely matching ids, so a tautology is not mistaken for a symmetry declaration.

## Recommended test, whichever way #126 goes

A canary that parses the two axioms above and asserts (a) they either parse or are surfaced in `dropped`, and (b) neither produces a symmetry or inverse-pair record. If the parser still rejects them the canary documents that as the current reason for safety, and it flips to a real guard the moment the parser changes — rather than the protection silently evaporating.

Related: #128 (whose fix widens the blast radius of a wrong symmetry record, since `role_matches` then reverses chain legs on it) and #144 (the sibling polarity guard, in the opposite direction — there an inverse-polarity symmetry axiom is wrongly *ignored*).

micheldumontier:
Fixes #145. Stacked on #150 (both touch `build_role_hierarchy`); merge that first or expect a trivial conflict.

## The hazard

`InverseObjectProperties(a, b)` asserts `a ≡ b⁻`. Two consumers compared role ids and threw polarity away — `collect_inverse_pairs` (which feeds the tableau's `declare_inverse_pair`) and the `mark_symmetric` arm of `build_role_hierarchy`:

| axiom | actually asserts | old reading |
| --- | --- | --- |
| `InverseObjectProperties(p, p)` | `p ≡ p⁻` — symmetric | symmetric ✔ |
| `InverseObjectProperties(p, ObjectInverseOf(p))` | `p ≡ (p⁻)⁻` — **tautology** | symmetric ✘ |
| `InverseObjectProperties(p, ObjectInverseOf(q))` | `p ≡ q` — role **equivalence** | `p ≡ q⁻` ✘ |

Row two invents a symmetry nobody declared. That is a false-**positive** source, and a widening one since #133 taught `role_matches` to walk chain legs backwards on a symmetry record.

Both sites now require **matching polarity**. Dropping the mixed case is an under-approximation — a role equivalence goes unrecorded — and therefore sound; recording a relationship the axiom does not assert is not.

## Why fix something unreachable

horned-owl rejects `ObjectInverseOf` inside `InverseObjectProperties`, and **that rejection is the only reason this is latent**. Konclude and HermiT both accept the construct, so it is a rustdl-side parser gap, not invalid OWL — the protection is an accident of a dependency rather than a property of this code.

I verified the horned-owl 1.4 → 3.0 upgrade (#126) did **not** open the door — the construct is still rejected on current `main`. But that was luck, not design: a parser bump could have turned a latent FP live with nobody touching the reasoner, and it would have surfaced as an unexplained false positive in an area where nothing changed.

## The test pins the reason, not just the behaviour

`crates/owl-dl-reasoner/tests/inverse_polarity_guards.rs`:

- **A tripwire** asserting the parser still rejects the construct, with instructions for when it flips: check the guards first, *then* replace the tripwire with the real behavioural assertions. So the protection cannot evaporate unnoticed — a future parser accepting it becomes a documentation event rather than an incident, since the guards are now written correctly regardless.
- **Two controls for the tightening itself**: a named self-inverse pair `(p, p)` must still yield symmetry, and a distinct named inverse pair `(co, s)` must still be an inverse pair. Without these, tightening "ids match" to "ids match AND polarity matches" could silently over-reject and nothing would notice.

## Verification

fmt, clippy `-D warnings`, and the full test suite green. Corpus soundness net running at time of writing; I will post the `[fp0]` manifest here.


micheldumontier:
Corpus soundness net complete — every fixture FP=0/MISSED=0, unchanged by this tightening:

| fixture | closure | |
|---|---|---|
| sio | 8904 = 8904 | FP=0 MISSED=0 |
| wine | 653 = 653 | FP=0 MISSED=0 |
| ro | 158 = 158 | FP=0 MISSED=0 |
| sulo | 51 = 51 | FP=0 MISSED=0 |
| pizza | 112 = 112 | FP=0 MISSED=0 |

Expected — the mixed-polarity case this guards against is unreachable through the parser, so the corpus exercises only the matching-polarity rows, which the two controls in the new test cover directly.

PR #158, opened 2026-09-14
test(hyper): pin the sub-role edge / super-role trigger gap (#128)

micheldumontier:
An adversarial review of the #128 measurement work raised this as an **unverified**
side observation. This PR verifies it and bounds it. Test only.

## The gap is real at the engine level

`index_one_clause` files a Horn clause under `role_trigger[` its body atom's **own**
role `]`; the `Event::Edge` dispatch looks up the **edge's** own role. So an `R`-edge
never wakes a clause whose body wants super-role `S` — even though `role_matches`,
once the hierarchy is threaded, would happily accept that edge.

`subrole_edge_trigger.rs` drives `HyperEngine` directly: correct answer `Unsat`,
engine says `Sat`. A control (no hierarchy → `Sat` is genuinely correct, since `R`
does not satisfy `S`) is included so the pin cannot pass for the wrong reason.

## It is NOT a live bug — three attempts to provoke it failed

| attempt | result |
|---|---|
| asserted ABox edge (`ObjectPropertyAssertion` + `ObjectPropertyDomain`) | correct (inconsistent), matches Konclude |
| generated TBox edge (`A ⊑ ∃R.⊤`) | correct (`A ⊑ C`), matches Konclude |
| disjunction-forced, intended to defeat the saturator | correct, matches Konclude |

The banner says why: **`subsumption: saturation=2 tableau=0`**. `ObjectPropertyDomain(S, C)`
is EL-expressible (`∃S.⊤ ⊑ C`), so the EL saturator answers it and the wedge is never
asked. A tableau-level test was required precisely because the pipeline never gets there.

That protection is an **accident of the saturator's coverage**, not a property of the
wedge — the same shape as #145, where a horned-owl parser gap was the only thing keeping
an unsound reading unreachable. So it is pinned rather than ignored (per
`docs/2026-08-18-ignored-sentinels-went-stale-unobserved.md`): if the covering mechanism
changes, this fails loudly instead of turning into a silent missed subsumption.

## The fix, if it is ever needed

Recorded in the test: `index_one_clause` already receives the hierarchy, so a role-atom
body on `r` can additionally be filed under every `h.sub_roles(r.role_id())` at matching
polarity. That widens the index — more clauses woken per edge, which is exactly the
Layer A cost profile — so it needs measuring, not assuming. Deliberately not done here.

Relevant to any Layer A default flip: turning that flag on is what makes `role_matches`
accept these edges in the first place.


micheldumontier:
Superseded by #159, which **closes** this gap rather than pinning it.

I opened this PR arguing the gap was unreachable and therefore not worth ~4% wall. That reasoning was wrong on both halves. I had established only that three attempts failed to provoke it — absence of evidence, not evidence of absence — and I weighed a few percent of wall against a known-wrong answer in a reasoner, which is the wrong trade. Corpus silence across nine fixtures is not a completeness proof.

Three further attempts after this PR also failed to reproduce it end-to-end (`ObjectPropertyRange`, a branch-forced successor, and `∀S.C` with the edge arriving after the label), all answered by the EL saturator with `tableau=0`. That still does not establish unreachability — it establishes that the saturator covers the shapes I could think of, which is the #145 pattern of protection-by-accident.

The test survives in #159 as a positive assertion, with its control and a mutation check.

PR #159, opened 2026-09-14
fix(hyper): a sub-role edge must wake a super-role-body clause (#128)

micheldumontier:
**A correctness fix.** Supersedes #158, which pinned this as a known gap on reasoning
I now think was wrong.

## The bug

`index_one_clause` files a Horn clause under `role_trigger[` its body atom's **own**
role `]`; the `Event::Edge` dispatch looks up the **edge's** role. So an `R`-edge never
woke a clause whose body wants super-role `S` (with `R ⊑ S`) — even though
`role_matches`, once the hierarchy is threaded, accepts that edge. It was never asked.
The engine returns `Sat` where `Unsat` is correct.

Role bodies are now filed under every sub-role as well. Trigger keys are polarity-free
(`role_id_index` maps `Named(x)` and `Inverse(x)` alike), so only the role id matters
here — `role_matches` still enforces polarity at match time. The back-trigger (HF3 chain
second leg) gets the same treatment for the same reason.

## Why ship it when no end-to-end case reproduces

Six attempts failed to provoke it through the pipeline:

| attempt | outcome |
|---|---|
| asserted ABox edge (`ObjectPropertyAssertion` + domain) | correct, matches Konclude |
| generated TBox edge (`A ⊑ ∃R.⊤`) | correct |
| disjunction-forced, to defeat the saturator | correct |
| `ObjectPropertyRange` (a `∀`, outside EL) | correct |
| branch-forced successor (`∃R.(D ⊔ E)`) | correct |
| `A ⊑ ∀S.C` + `A ⊑ ∃R.D` (edge arrives after label) | correct |

All showed `tableau=0` — the EL saturator got there first every time.

**That is not a reason to leave it.** "I could not construct a reachable case in six
tries" is a fact about my fixtures, not about the space of OWL ontologies. The wedge is
reachable in general — wine, galen and the ORE ontologies all show non-zero tableau
subsumption calls — and without this fix the reasoner's correctness on those paths rests
on the saturator *happening* to cover a gap nobody designed it to cover. That is exactly
the #145 shape (a horned-owl parser gap was the only thing keeping an unsound reading
unreachable), and #145 was fixed rather than pinned.

## Cost, measured

| workload | base | fixed | Δ |
|---|---|---|---|
| wine (Layer A off, 5 reps) | 7.59 s | 7.92 s | +4.4%, ranges overlap |
| galen (Layer A off, 3 reps) | 3.41 s | 3.57 s | ~+5% |
| sio / ro / pizza | — | — | flat |
| `ore_ont_16372` (Layer A **on**) | 31.1 s | 30.8 s | flat |
| `ore_ont_9890` (Layer A **on**) | 42.4 s | 42.4 s | flat |

The cost lands on the Layer-A-**off** path and is zero on Layer-A-**on** — the opposite
of my prediction. Cause: `ConsistencyCache::build_seeded_engine` threads the hierarchy
**ungated**, so ABox-bearing ontologies get the widened index regardless of the flag.
Wine and galen have ABoxes; sio/ro/pizza effectively don't, and they are flat.

## Verification

- `cargo test` — 0 failed.
- Soundness net — **9 VERIFIED, 20 closure measurements, all FP=0 MISSED=0**, every
  closure count byte-identical to baseline (galen 28007, sio 8904, ore-10908 6001,
  wine 653, ro 158, ore-15672 142, pizza 112, sulo 51). No answer changes.
- The test is **mutation-checked**: reverting just the widening flips it back to `Sat`.
  A control (no hierarchy ⇒ `Sat` is genuinely correct) ensures it cannot pass for the
  wrong reason.

## What I got wrong first time

I originally proposed pinning this and not fixing it, on the grounds that it was
"unreachable" and cost ~4%. Two errors: I had established only that *three attempts
failed*, which is absence of evidence; and I weighed a few percent of wall against a
known-wrong answer in a reasoner, which is the wrong trade. Corpus silence across nine
fixtures is not a completeness proof.


micheldumontier:
## Update: the first commit was incomplete, and an audit caught it

`inverse_first_trigger` has the **identical** defect — keyed by the clause's own role,
looked up under the *edge's* role at dispatch. So a clause whose first leg is an inverse
or symmetric atom on `S` was still never woken at the target by an `R`-edge with `R ⊑ S`.
I had widened two of the three role-keyed tables and missed the third.

That shape is not exotic: `S⁻(x,y) → Q(x)` is what `ObjectPropertyRange(S, Q)` becomes at
the far endpoint.

### The audit

Stated the invariant the original bug violated — *every edge `role_matches` would accept
for a body atom must wake that clause* — and checked each table against it:

| path | hierarchy-aware at dispatch? |
|---|---|
| `role_trigger` | **was broken** — fixed, first commit |
| `role_back_trigger` | **was broken** — fixed, first commit |
| `inverse_first_trigger` | **was broken** — missed, fixed in the second commit |
| `at_most_ones` / `distinct_role_succ` | already correct — both consult `role_matches` with the hierarchy by design |
| `x_trigger` / `succ_trigger` | N/A — exact class keys, no hierarchy |

Cardinality was already right, so sub-roles *were* considered there. It was specifically
the role-keyed trigger family that was maintained case by case — inverse and symmetric
handled via `inverse_first_trigger`, sub-roles not handled at all.

### Re-measured with both widenings (5 reps, clean release binaries)

| fixture | base | both fixes | Δ |
|---|---|---|---|
| wine | 7.44 s | 7.86 s | **+5.7%** |
| galen | 3.62 s | 3.56 s | −1.7% |
| sio | 1.14 s | 1.13 s | −1.4% |
| ro | 0.61 s | 0.63 s | +3.3% (0.02 s absolute) |

**Correction to the original PR body:** it reported galen at ~+5% slower from 3 reps. At
5 reps galen is −1.7%, i.e. that figure was noise. Wine is the only fixture with a
consistent cost.

### Gate, re-run with both fixes

`cargo test` 0 failed. Soundness net **9 VERIFIED, 20 closure measurements, all FP=0
MISSED=0**, every closure count byte-identical to baseline.

Both widenings are mutation-checked independently: reverting either one alone flips its
own test back to `Sat`, and each has a control asserting that *without* the hierarchy
`Sat` is the correct answer, so neither can pass for the wrong reason.

micheldumontier:
## ORE validation (610 oracle-backed ontologies) — 0 attributable differences

The nine-fixture corpus was not adequate evidence for a change to the clause **trigger
index**, which decides what fires at all. Paired sweep over every ORE pool ontology that
has a Konclude oracle, base vs fixed in the same worker slot back to back:

| | count |
|---|---|
| SAME | 601 |
| DIFF (first pass, 8-way parallel, 600 s cap) | 9 |
| **DIFF surviving serial recheck** | **0** |

**No FP introduced anywhere. No MISSED regression.**

### All 9 first-pass DIFFs were measurement artifacts

They initially looked alarming and uniformly against the fix — `ore_ont_7499` MISSED
0 → 13, four large ontologies completing on base and timing out on fixed. Re-run
serially with a 1200 s cap, every one is SAME:

    ore_ont_7499   base 36145/36145 MISSED=0   fixed 36145/36145 MISSED=0   SAME
    ore_ont_205    base 38638/38638 MISSED=0   fixed 38638/38638 MISSED=0   SAME
    ore_ont_2302   base 507351 MISSED=0        fixed 507351 MISSED=0        SAME
    ore_ont_4782   base 144565 MISSED=0        fixed 144565 MISSED=0        SAME
    ore_ont_1199   base  56408 MISSED=0        fixed  56408 MISSED=0        SAME
    ore_ont_6726   base  48324 MISSED=0        fixed  48324 MISSED=0        SAME
    ore_ont_14879  base      0 MISSED=0        fixed      0 MISSED=0        SAME
    ore_ont_9662   base 9162/9164 MISSED=2     fixed 9162/9164 MISSED=2     SAME
    ore_ont_7532   base 10551/10553 MISSED=2   fixed 10551/10553 MISSED=2   SAME

### Wall cost on the large ORE ontologies: nil

Paired, serial, same slot:

| ontology | base | fixed | Δ |
|---|---|---|---|
| `ore_ont_2302` | 296.2 s | 299.8 s | +1.2% |
| `ore_ont_4782` | 185.3 s | 185.5 s | +0.1% |
| `ore_ont_1199` | 135.9 s | 138.1 s | +1.6% |
| `ore_ont_6726` | 209.0 s | 199.8 s | −4.4% |
| `ore_ont_205` | 337.9 s | 333.7 s | −1.3% |
| `ore_ont_7499` | 290.8 s | 294.7 s | +1.3% |

Mean ≈ −0.2%. The first-pass timeouts were cap-plus-load: these run 135–340 s serially,
so 8-way contention against a 600 s cap flips whichever arm draws the busier moment.

**Net cost across everything measured: wine +5.7% (5 reps), everything else within noise.**
Wine is the ABox-heavy fixture, consistent with `ConsistencyCache::build_seeded_engine`
threading the hierarchy ungated — worth fixing separately.

## Harness finding worth recording independently

**The ORE closure-diff harness is not reproducible at `RUSTDL_TEST_PAIR_MS=200` under
parallelism.** The spurious-DIFF rate here was 9/610 ≈ **1.5%**, and it is not confined to
the arm under test — `ore_ont_205`'s BASE arm gave 38636/38638 MISSED=2 in the parallel
sweep and 38638/38638 MISSED=0 serially. Pairing the arms protects against drift across a
sweep but not against the two runs of a pair landing on differently-loaded moments.

Practical rule: a sweep surfacing a handful of DIFFs is indistinguishable from noise until
each is rechecked serially. I had begun writing up a causal story for these nine — extra
firing exhausting the per-pair budget and defaulting pairs to not-subsumed — which was
entirely wrong.

